Issue #2, opened 2026-06-22
index sync/build ignore DB-registered KBs (get_config_and_db skips _register_db_kbs)


markramm:
## Summary

`pyrite index sync` (and the other `index_commands.py` entry points) operate on a config that is **missing any KB added via `pyrite kb add`**, so indexing a DB-registered KB produces **0 entries** even though `pyrite kb list` shows it.

## Root cause

`pyrite/cli/index_commands.py` builds its context via `get_config_and_db()`:

```python
# index_commands.py:42, 142, 197
config, db = get_config_and_db()
```

But `get_config_and_db()` (`pyrite/cli/context.py:85`) only does `load_config()` — it never calls `_register_db_kbs()`:

```python
def get_config_and_db() -> tuple[PyriteConfig, PyriteDB]:
    config = load_config()
    db = PyriteDB(config.settings.index_path)
    return config, db          # <-- no _register_db_kbs(config, db)
```

By contrast, the context-manager path `_init_base()` (`context.py:44`) DOES merge them:

```python
def _init_base():
    config = load_config()
    db = PyriteDB(config.settings.index_path)
    _register_db_kbs(config, db)   # <-- DB-registered KBs merged in
    ...
```

So KBs registered through `pyrite kb add` (which live in the DB, not the YAML config) are invisible to `index sync`.

## Repro

```bash
pyrite kb add /path/to/some-kb -n some-kb        # registers into the DB
pyrite kb list                                   # some-kb shows up
pyrite index sync -k some-kb                      # "Updated: 0" — indexes nothing
pyrite search "<known term>" -k some-kb           # 0 results
```

## Impact

Hit in a cloud/headless setup where KBs are registered via `kb add` at runtime (no YAML config edit). The only workaround was to hand-edit the YAML config to add the KB path so `load_config()` picks it up. This makes `kb add` + `index sync` unusable together.

## Suggested fix

Make `get_config_and_db()` call `_register_db_kbs(config, db)` before returning (matching `_init_base()`), or migrate `index_commands.py` to use `cli_db_context()` / a context that performs the merge.

markramm:
Fixed on `dev` in commit f193ef7.

**Root cause** was two-part:
1. `get_config_and_db()` / `cli_db_context()` never merged DB-registered KBs (your diagnosis — they only called `load_config()`). Now they call `_register_db_kbs()`, fixing `index sync -k <kb>`.
[...]

Issue #1, opened 2026-04-03
Link commands fail with 'KB not found' for newly registered KBs


markramm:
## Bug Description

The `kb link` and `kb links bulk-create` commands fail with `Error: KB not found` for a newly registered and indexed KB, even though `kb search`, `kb get`, `kb list`, and `kb tags` all work correctly for the same KB.

## Steps to Reproduce

1. Create a new KB directory with kb.yaml and entries
2. Register it: `kb kb add /path/to/kb -n my-kb -t network-investigation -d "description"`
3. Index it: `kb kb reindex -k my-kb` → succeeds, shows entries added
4. Verify search works: `kb search "test" -k my-kb` → succeeds, returns results
5. Verify get works: `kb get entry-id -k my-kb` → succeeds
6. Try to create a link: `kb link source-id target-id -k my-kb -r related_to` → **fails with "KB not found"**
7. Try bulk links: `kb links bulk-create file.yaml -k my-kb` → **same error**
8. Also fails: `kb kb validate my-kb` → "KB not found"
9. Also fails: `kb orient -k my-kb` → "KB not found"

## Expected Behavior

All commands should find the KB consistently once it's registered and indexed.

## Actual Behavior

Search index commands (search, get, tags, list-entries, reindex) work fine.
Link commands, validate, and orient all fail with "KB not found".

This suggests the link subsystem uses a different registry/DB lookup path than the search subsystem.

## Environment

- Pyrite version: latest (as of 2026-04-03)
- KB type: network-investigation
- Entries: 106
- Registration method: `kb kb add` followed by `kb kb reindex`

## Workaround Attempted

Removed and re-added the KB — same result. The search index is correctly populated but the link/validate/orient subsystems can't find the KB.

## Affected Commands

- `kb link` ❌
- `kb links bulk-create` ❌  
- `kb kb validate` ❌
- `kb orient` ❌
- `kb search` ✅
- `kb get` ✅
- `kb tags` ✅
- `kb list-entries` ✅
- `kb kb reindex` ✅
- `kb kb list` ✅ (shows KB in registry)

## Investigation Notes

The main Pyrite DB at `~/.pyrite/pyrite.db` appears empty (`.tables` and `.schema` return nothing), while the search index is populated. The link system may be looking for entries in the main DB rather than the search index, and the main DB may not have been provisioned for this KB type.

markramm:
## Additional Finding

[...]

PR #4, opened 2026-09-09
Fix: KB created via POST /api/kbs is invisible to entry creation until restart

AsyncLegs:
## Summary

Related to (but narrower than) #collapse-kb-registry-to-one-source-of-truth: creating a KB via `POST /api/kbs` (or the `KBRegistryService.add_kb()` path generally) writes it to the DB successfully, but the very next `POST /api/entries` against that KB fails with:

```json
{"detail":{"code":"KB_NOT_FOUND","message":"KB 'homelab' not found"}}
```

...until the server process is restarted, at which point it works fine.

## Root cause

`PyriteConfig._db_kb_cache` — the fallback lookup `get_kb()`/`all_kbs()` use for DB-registered (as opposed to `config.yaml`-declared) KBs — is only populated **once**: `create_app()`'s `_app_get_db()` closure calls `db.merge_registered_kbs(cfg)` lazily on the first request that touches the DB, then caches `application.state.pyrite_db` and never calls it again for the lifetime of the process.

`KBRegistryService.add_kb()` writes the new KB straight to the `kb` table via `db.register_kb()`, but never touches `config._db_kb_cache`. Since `KBService.get_kb()` and `PyriteConfig.get_kb()` both resolve DB-only KBs through that in-memory cache (not a live DB query), the newly-created KB is invisible to every KB-name lookup that goes through `config` — including entry creation — for the rest of the process's life, even though it's already committed to the database. A CLI invocation (`pyrite kb add`) works around this by nature, since each CLI run calls `load_config()` fresh and re-merges from the DB.

## Fix

`add_kb()` now calls `self.config.register_db_kbs([...])` with the newly-registered KB immediately after the DB write, so the in-memory cache reflects it in the same process — no restart needed.

## Verification

Reproduced against a live deployment: `POST /api/kbs` to create a KB, then immediately `POST /api/entries` against it → `404 KB_NOT_FOUND`. Only a full pod restart made the KB usable. With this fix, built and deployed, the same create-KB-then-immediately-write-an-entry sequence succeeds without any restart.

## Test plan

- [ ] `POST /api/kbs` to create a new KB, then immediately `POST /api/entries` against it in the same process — should succeed, not 404
- [ ] Existing `tests/test_kb_registry_service.py` / `tests/test_services.py` coverage for `add_kb()` still passes

### DIFF
--- a/pyrite/services/kb_registry_service.py
+++ b/pyrite/services/kb_registry_service.py
@@ -128,6 +128,27 @@ class KBRegistryService:
             description=description,
             source="user",
         )
+
+        # config._db_kb_cache is only populated once, on first DB access
+        # (create_app()'s _app_get_db() calls db.merge_registered_kbs()
+        # lazily and then never again). Without this, a KB registered here
+        # is invisible to config.get_kb()/all_kbs() -- and therefore to
+        # KBService.get_kb() and entry creation -- for the rest of this
+        # process's life, even though it's already committed to the DB.
+        # Update the in-memory cache immediately so it's usable in the same
+        # request/process, not just after a restart.
+        self.config.register_db_kbs(
+            [
+                {
+                    "name": name,
+                    "path": str(resolved),
+                    "kb_type": kb_type,
+                    "description": description,
+                    "default_role": None,
+                }
+            ]
+        )
+
         return self.get_kb(name)  # type: ignore[return-value]
 
     def _apply_preset(self, kb_path: Path, name: str, kb_type: str, description: str) -> None:


markramm:
Merged, in [v0.24.1](https://github.com/markramm/pyrite/releases/tag/v0.24.1). Thanks for the root-cause write-up; it was better than the ticket I had.

You were right that this is narrower than `collapse-kb-registry-to-one-source-of-truth`, but it turned out to be the fourth time this class of bug has shown up (issues #1 and #2 and one earlier commit were the same cache going stale), so I've raised that item to high priority. Your fix is the right patch for now; the real fix is one registry with no cache to refresh.

One thing I owe you: there's no test that `add_kb()` refreshes the cache, so this could regress silently. I verified the flow by hand today (create KB over REST, then create an entry in it, no restart) and that's now a required step in the release checklist, but a real test is tracked with the integration-test work above.

AsyncLegs:
Its my pleasure 😇

Issue #78, opened 2026-09-18
Flaky teardown: test_api_tiers three_key_client TemporaryDirectory cleanup races under -n auto


markramm:
`tests/test_api_tiers.py::TestTierHierarchy::test_admin_key_can_access_all_tiers` intermittently ERRORs at teardown under `pytest -n auto` (roughly 1 run in 3 on an 4200-test full-suite run; never when the file is run alone).

```
__ ERROR at teardown of TestTierHierarchy.test_admin_key_can_access_all_tiers __
tests/test_api_tiers.py:166: in three_key_client
    with tempfile.TemporaryDirectory() as d:
...
E   OSError: [Errno 66] Directory not empty: '/var/folders/.../tmpvn_3dc8h'
```

The class-scoped `three_key_client` fixture (tests/test_api_tiers.py:163-177) closes `client.pyrite_db` in its `finally`, but something in the temp dir is still being written when `TemporaryDirectory.__exit__` runs `rmtree` — most likely SQLite WAL/shm sidecar files recreated by a connection the app still holds, or a background writer that has not drained. Under load the window widens, which is why it only shows under `-n auto`.

No test fails — all 4244 pass — but the error makes the suite's exit status non-zero, so it can redden CI and a pre-push run for reasons unrelated to the change being pushed.

Found while working the milestone-0.24.2 CLI theme (#46/#18/#21) on `fix/cli-write-and-report-correctness`; confirmed pre-existing by reproducing with that branch's changes stashed. Not fixed there, per ADR-0033.

Likely fix: close/dispose every engine and connection (not just `pyrite_db`) before leaving the `with`, or let pytest's `tmp_path_factory` own the directory instead of `tempfile.TemporaryDirectory`, so cleanup failures do not raise.

markramm:
Duplicate of #55 — same fixture (`tests/test_api_tiers.py` `three_key_client`), same `OSError: Directory not empty` at teardown, same `-n auto` condition. Filed from the #69 worktree without sight of #55.

Covered by **PR #81** (the quality theme: test teardown races), which fixes #55 together with the sibling `PyriteDB`-connection-leak backlog item — the two are the same failure shape from two directions. Leaving this open until #81 lands, then closing as a duplicate.

Found by the tick-3 architect groom, verified against the code.

markramm:
Conductor review note (tick 2026-09-18, reviewing branch
`fix/cli-write-and-report-correctness` / PR #69, which filed this issue).

[...]

PR #81, opened 2026-09-18
quality: test teardown races — no test leaves a writer running inside a temp dir

### LINKED ISSUE #55: tests: test_api_tiers teardown errors under -n auto (temp dir removed while an index worker thread is still writing)

markramm:
A test run no longer fails because something was still writing inside a directory pytest was deleting. This removes the one known intermittent red on `dev` — the failure that teaches every agent to ignore a red — and it does so by closing the connections and joining the threads, not by making the error quieter.

1. **`PyriteDB` is a context manager.** `with PyriteDB(path) as db:` closes on block exit and on exception (`pyrite/storage/connection.py`). An unclosed SQLite connection in WAL mode recreates its `-wal`/`-shm` files while `rmtree` walks the directory — one of the two failure shapes.

2. **One shared `make_client(tmp_path, **settings)` fixture replaces four hand-rolled helpers** (`tests/conftest.py`), owning every DB opened on its behalf and the app's `IndexWorker`. Note the diff is **+369/−301**: this replaces duplication rather than layering over it. Teardown order is load-bearing and commented: join the worker threads *first*, then close the connections they write through, both before the directory goes.

3. **Fixes #55 — the index-worker race — in the test, not the product.** `test_api_tiers.py`'s class-scoped fixtures build their own app via `create_app()`, which lazily creates an `IndexWorker` on the first `/api/index/sync`; nothing joined that thread before the class-scoped temp dir was removed. `IndexWorker.wait_for_idle()` and its thread tracking already existed — the gap was that the test's own app-building code never called them. No change to `pyrite/services/index_worker.py` was needed or made.

4. **A cold read found the fixture's central claim was false, and it is now true.** `create_app()` eagerly seeds the KB registry, which calls its own `_app_get_db()` and opens a **second** `PyriteDB` on the same file, parked on `application.state.pyrite_db`. The `dependency_overrides` redirect DI but do not close it; some routes read it directly rather than through DI (the export path in `pyrite/server/endpoints/kbs.py`); and **nothing in `pyrite/` ever closes it**. So the leak this theme is named after was still present — relocated, and invisible only because the branch also moved to `tmp_path`, which pytest does not delete during a run. A *default* (`tmp_path_retention_policy`) was doing the work the code claimed to do. The fixture now tracks and closes that connection too.

5. **One failing close can no longer leak the rest.** Both teardown loops — `make_client`'s and `test_api_tiers`' four-client fixture — guard each join and each close individually and re-raise the first error, so a real failure stays loud. Previously one raise skipped every close after it: the same class of bug the fixture exists to prevent.

6. **The ticket's criterion 1 is now actually met.** Three sites still matched its exact pattern and were not in its nine-file list: `test_collections.py`, `test_collection_query.py`, and `test_admin_cli.py` — the last constructed inline inside a `patch()` call and never bound to a name, so it could not be closed at all.

**Evidence.**
- Pre-fix reproducer: `pytest tests/test_api_tiers.py tests/test_index_worker.py -n auto`, 6 runs, run 3 failed with `OSError: [Errno 66] Directory not empty` — ~1 in 6.
- Post-fix, from three independent sources: the worker's 10-run targeted pass and 10-run full-suite pass; a detached full-suite run (4222 passed); and the conductor's own 9 reproducer runs. **Zero teardown errors in every run.**
- Final full suite on this branch: **4225 passed, 68 skipped, 7m15s, zero teardown errors.** The count rises from 4223 by exactly the two context-manager tests added; nothing deleted, skipped or weakened.
- `tests/test_make_client_closes_every_connection.py` verifies red: with the app-state tracking removed it fails with `WAL/SHM still live after teardown: ['index.db', 'index.db-shm', 'index.db-wal']`. It asserts on the *connection*, not on whether a directory removal happened to survive, because the retention policy is a default a project can change — and it guards its own premise, so it fails loudly rather than passing vacuously if `create_app` ever stops opening the second connection.

**Filed, not fixed.** #88 — `test_index_worker.py`'s `test_active_jobs_filters` and `test_duplicate_sync_returns_same_id` assume `submit_sync()`'s thread has *not* finished by the next line, which only holds while the job is slow; under `-n auto` on a loaded machine a near-empty KB's sync wins the race. Pre-existing, reproduces unpatched, and a different defect from this theme's (a same-thread assertion racing the job's own completion, not a teardown race). It is the occasional non-teardown failure visible in the runs above.

**Known trade-offs the reviewer should see**, all from the cold read, none blocking:
- `close()` is not terminal — SQLAlchemy silently re-opens from the pool, so a caller holding a reference past the `with` block can resurrect the WAL files. `__exit__` also does not guard re-entry. A `_closed` flag would make the new public API mean exactly what its docstring says.
- `IndexWorker._threads` is never pruned, so `wait_for_idle`'s 10 s timeout is per-thread and cumulative rather than a teardown budget.
- The new context manager has **zero adoption outside `tests/`**: `pyrite/cli/context.py` still hand-rolls `try/finally: db.close()` three times, which is exactly where it would earn its keep next.
- Five fixtures in `test_api_tiers.py` gained `@classmethod` above `@pytest.fixture` to clear pytest's class-scoped-fixture deprecation. It works and teardown was verified to run, but the decorator ordering is not documented as supported and the project pins only `pytest>=8.0.0`.

Fixes #55.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


markramm:
Cross-posting measurements from another conductor session's review of PR #69,
because they bear directly on this theme's `test_api_tiers` half (#55, and its
duplicate #78 filed from the other direction).

**The flake is load/timing-sensitive, and the exposure is branch-dependent.**
Same command (`pytest tests/test_api_tiers.py tests/ -n auto -q`), alternating
trees, on one machine:

| tree | full-suite runs | `test_admin_key_can_access_all_tiers` teardown ERRORs |
|---|---|---|
| clean `dev` (4e4368f) | 4 | **0** |
| `fix/cli-write-and-report-correctness` (PR #69) | 4 | **2** (a 5th targeted run reproduced immediately) |

So "one run in three" is not a fixed rate — it depends on what else is in the
suite and how fast it runs. A fix verified by "it stopped failing for me" is
therefore not verified at all; **the acceptance needs a deterministic
reproduction**, not N green runs. Suggestion, if it helps: make the writer's
liveness observable and assert on it directly (e.g. after the fixture's
teardown, assert no `-wal`/`-shm` remain and the worker thread is not alive),
so the test proves the invariant rather than sampling a race.

**What I confirmed about the mechanism**, which matches this theme's framing:

- The fixture at `tests/test_api_tiers.py:163-177` (`three_key_client`,
  class-scoped) calls `client.pyrite_db.close()`, and **`index.db-shm` and
  `index.db-wal` still remain** in the `TemporaryDirectory` afterwards. I
  probed this directly and the leftover set is **identical on `dev` and on the
  branch**:
  ```
  index.db  index.db-shm  index.db-wal  kb/notes/t.md
  ```
  So `close()` is not sufficient to checkpoint/remove the WAL companions —
  consistent with your "unclosed connection in WAL mode recreating `-wal`/`-shm`"
  hypothesis, and it reproduces with a single `POST /api/entries` then `close()`.
- The error is always `OSError: [Errno 66] Directory not empty` from
  `TemporaryDirectory.cleanup()` → `shutil.rmtree` → `os.rmdir`, i.e. a file
  appearing *after* rmtree walked the directory — a writer still live at
  teardown, not merely a file left behind.

**One thing to be careful of in triage.** #78 asserts this is pre-existing and
unrelated to PR #69, "confirmed by reproducing it with this branch's changes
stashed". The numbers above do not support that claim. The *race* is certainly
pre-existing (identical leftovers on both trees, and #55 predates #69), but
something on #69 takes it from never-observed to ~50% under `-n auto`, and I
could not find the mechanism: the obvious suspect is ruled out, as that
branch's new write-path work is not a slowdown (200 load+save round trips:
0.182s on the branch vs 0.214s on `dev`). Worth knowing while you work here,
because if your fix is validated only against `dev` it may look complete while
#69 still reddens the suite. Validating the fix against **both** trees would
settle it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #64, opened 2026-09-18
Dangling wikilinks come back as title: null with no resolved flag (kb_get outlinks, cascade_network)


markramm:
Source: read-tier hallway test, 2026-09-18, cold-start session over the MCP tools (54 KBs, 27,135 entries). Every error message quoted verbatim from that session.

```json
{"id":"feedback_party_filing_not_authoritative_for_another_docket","kb_name":"cascade-research","title":null,"entry_type":null,"relation":"wikilink"}
```

The caller must infer "null title ⇒ target does not exist". In `cascade_network(thiel-peter)`, 5 of 35 outlinks were nulls. The corpus has a method entry on exactly this ambiguity (a dangling wikilink is usually a cross-KB gap: "12 of 13 'missing' entities already existed").

Fix: `resolved: false` on unresolved links, and `resolved_in_kb: "<other-kb>"` when the target id exists in another registered KB. Acceptance: a two-KB fixture; the REST link endpoints return the same fields.

markramm:
## Triage 2026-09-20

**State: ready** — and it should be groomed **with #124**, which is the write-side half of the same missing concept.

The two issues are one idea seen from both ends: #124 adds `allow_dangling` and a `resolved` flag when a link is **created**; this adds `resolved: false` (and `resolved_in_kb`) when a link is **read**. Shipping either alone leaves the API half-honest — a caller can create a forward reference it cannot then distinguish on read, or read a `resolved` flag that no write path ever sets deliberately. **Recommend one theme, closing both**, and say so in the PR title.

**Acceptance** (from this body): unresolved links carry `resolved: false`; `resolved_in_kb: "<other-kb>"` when the target id exists in another **registered and readable** KB; a two-KB fixture; **the REST link endpoints return the same fields** as MCP. Plus #124's half: `allow_dangling: bool = False` on `kb_service.add_link` and in the `kb_link` tool schema; `add_link` returns `{"resolved": bool}`; `_kb_link` passes it back; and #124's named defect — the duplicate-link early return must **re-check that the existing target still exists** rather than reporting `resolved: True` blindly.

[...]

PR #108, opened 2026-09-18
fix: validate target entry in kb_link to prevent dangling links (#97)

### LINKED ISSUE #97: kb_link validates the source entry but not the target — silently creates dangling links (mechanism behind #64)
Source: write-tier hallway test, 2026-09-18, same cold-start MCP session as #56–#68, against two disposable sandbox KBs (`hallway-sandbox-generic`, plain schema; `hallway-sandbox-investigation`, `network-investigation` schema) registered with `kb_registry_add` and removed afterwards. Never against a working corpus. Errors verbatim.

```
kb_link(source_id="round-trip-test-does-not-exist-yet", source_kb="hallway-sandbox-generic", target_id="zzz-nonexistent-target-entry")
→ {"error": "Entry not found: round-trip-test-does-not-exist-yet", "error_code": "LINK_FAILED", "retryable": true}

kb_link(source_id="sandbox-round-trip-test-entry", source_kb="hallway-sandbox-generic", target_id="zzz-nonexistent-target-entry", relation="related_to")
→ {"linked": true, ...}
kb_get(entry_id="sandbox-round-trip-test-entry", ...)
→ "outlinks": [{"id": "zzz-nonexistent-target-entry", "kb_name": "hallway-sandbox-generic", "title": null, "entry_type": null, "relation": "related_to"}]
```

The source is validated; the target is not. This is the mechanical origin of the `title: null` dangling outlinks in #64 (5 of 35 on one hub node). Also: `retryable: true` on a not-found error is wrong the same way as #57 — a nonexistent id does not start existing on retry.

## Acceptance

- `kb_link` validates the target exists in the target KB (or, when the caller wants a forward reference, requires `allow_dangling: true` and reports `resolved: false` per #64).
- `LINK_FAILED` for a missing entry is `retryable: false`.
- Tests for both sides.

fathirramadhan-web:
Closes #97

### Overview
Adds target entry validation in `kb_link` before creating links. If the target entry does not exist, raises an error instead of silently creating a dangling link.

### Changes
- `pyrite/services/kb_service.py`: Added target entry existence check
- `pyrite/server/mcp_server.py`: Updated error handling for invalid targets
- `pyrite/server/tool_schemas.py`: Schema updates
- `tests/test_mcp_server.py`: Added test coverage for dangling link prevention

### Verification
- Tested locally
- Linters/tests pass

Fixes #97


markramm:
## Conductor partial review (tick 5) — handing off to a dedicated reviewer

Context: a dedicated review agent is taking this PR over. Posting what I verified so it is not lost, and releasing my `in-review` claim. **No merge decision here — outside contributions are the maintainer's call.**

**CI has never run on this PR.** The workflow run for `565a6f5` is `conclusion: action_required` — held pending maintainer approval for a first-time contributor. Nothing here is machine-verified yet. Only the maintainer can approve it.

**This PR is currently `DIRTY`** (conflicts against `dev`) and needs a rebase before it can be assessed properly.

What is good and is genuinely ahead of the competing #116: the `allow_dangling` escape hatch with a `resolved` flag in the response is a thoughtful design — it distinguishes "you typo'd" from "forward reference I will fill in later", and reporting `resolved` back to the caller lets an agent notice. Returning a dict from `add_link` instead of `None` is a clean way to carry that. The test set is the broader of the two (target-missing, allow-dangling, resolved-true).

Two things the dedicated reviewer should look at closely:

1. **Validation reads the index, not the disk.** `self.db.get_entry(target_id, tkb)` is the SQLite index (`pyrite/storage/crud.py:22`). An entry that exists on disk but has not been `index sync`'d is invisible to it, so this will reject a link whose target is really present — a false negative on the common `pyrite create` → `pyrite link` path. #116 uses `KBRepository.load` (disk) and does not have this problem. This is the substantive difference between the two PRs and #116's description is right about it.

2. **The `.gitignore` hunk does not belong to this change.** It appends, under a comment reading `# Auto-added by github-engineer`, patterns for `*.obj`, `*.so`, `*.whl`, `nem_stats.c`, `*.exp`, `*.lib` — `nem_stats.c` in particular is from an unrelated project. This is tooling spillover and should be dropped from the PR whatever else is decided.

Also worth confirming: the duplicate-link early return now reports `{"resolved": True}` without checking whether the existing link's target still exists, so a pre-existing dangling link reports as resolved.

Conflict with #69: no textual conflict observed; #69 rewrites frontmatter emission in `Entry.to_markdown`, this touches `kb_service.add_link` ahead of the save. Assertions about what ends up stored should be re-run once #69 lands.

fathirramadhan-web:
and

markramm:
Thank you for this, and for tackling #97 — this is a real contribution and it made the product better.

The maintainer has now read both this and #116, which arrived independently for the same issue. The decision is to build on **#116**, on one specific technical point: it validates the target with `KBRepository.load` (the filesystem) where this PR validates with `self.db.get_entry` (the SQLite index, `pyrite/storage/crud.py:22`). An entry created on disk but not yet `index sync`'d is invisible to the index, so index-based validation rejects links whose target genuinely exists — and `pyrite create` immediately followed by `pyrite link` is the common agent path. That is the whole of it.

**That is not a judgement on the rest of this PR, which is in several ways the stronger of the two.** Specifically:

- **`allow_dangling` plus the `resolved` flag is the better design.** It distinguishes "you typo'd the target" from "forward reference I am about to create", and returning `resolved` to the caller lets an agent notice a link that did not land. #116 has no escape hatch at all. We have raised this on #116 and it will either be added there or filed as a follow-up **crediting this PR**.
- **Your test set is the broader one** — target-missing, allow-dangling, and resolved-true, where #116 covers the first only.
- Changing `add_link` to return a dict rather than `None` is the clean way to carry that information out.

Two small things for the record, in case you pick up the follow-up:

- The `.gitignore` hunk does not belong to this change — it appends `*.obj`, `*.so`, `*.whl`, `nem_stats.c`, `*.exp`, `*.lib` under a `# Auto-added by github-engineer` comment, and `nem_stats.c` is from an unrelated project. Tooling spillover; worth watching for.
- The duplicate-link early return now reports `{"resolved": True}` without re-checking the existing link's target, so a pre-existing dangling link would report as resolved.

Also worth knowing: CI never ran on this PR (`action_required`, held pending maintainer approval for first-time contributors), and the branch is currently `DIRTY` against `dev`. So none of the above comes from a failed check — it is a design comparison, not a quality verdict.

If you would like to carry the `allow_dangling` design forward as its own PR once #116 lands, we would welcome it and it would go in with your name on it. Thanks again.

markramm:
## Full review — PR #108

Thanks for this, @fathirramadhan-web, and sorry for the wait. This is a proper fix: you read the acceptance criteria on #97 closely, implemented all three of them, and wrote real tests. A second PR (#116) arrived for the same issue about twenty minutes apart, so this review also compares the two — that is on our issue tracker for handing the same ticket to two people, not on you.

A partial review from an earlier pass is above; this supersedes it. Everything below was verified by running it, not read off the diff.

### What I checked

- Read every hunk of `git diff origin/dev...HEAD`.
- Full suite on your head as-is: **4224 passed, 68 skipped, 0 failed** (139s, `pytest tests/ extensions/ -n auto`). Your branch is green on its own.
- Verify-red: reverted only the three implementation files to `dev` and re-ran your four tests. **All four fail without the fix**, each for the right reason (`retryable` still `True`; the dangling link still created; `KeyError: 'resolved'` twice). That is exactly the discipline we ask for and a lot of contributors skip it.
- `ruff check` — clean. `ruff format --check` — one finding, below.
- Merge state against `dev`, conflict by conflict.
- Behaviour probes for the index/disk question, the unregistered-target-KB path and the duplicate-link path, written as throwaway tests and run against both this PR and #116.
- Overlap with PR #69 (in flight, rewrites the entry save path).

### What is good, specifically

- **`allow_dangling` plus a `resolved` flag is the better design of the two PRs, and it is the one #97 actually asked for.** Acceptance criterion one says "or, when the caller wants a forward reference, requires `allow_dangling: true` and reports `resolved: false` per #64". #116 implements strict validation only; you implemented the escape hatch too. Distinguishing "you typo'd the id" from "forward reference I will fill in later" is the thing an agent building a KB incrementally actually needs, and reporting `resolved` back in the response is what lets it notice. Changing `add_link` to return a dict instead of `None` is a clean way to carry that without a second call.
- The schema entry in `tool_schemas.py` has a real description and an explicit default — an MCP client can discover the new argument without reading the source.
- Your test set is the broader of the two: target-missing, allow-dangling, resolved-true, and you tightened the existing `test_kb_link_not_found` to assert `retryable is False` rather than just `"error" in result`.

### What needs to change

**1. Validate against the files, not the index.** This is the one substantive problem, and it is the reason my recommendation lands where it does.

`self.db.get_entry(target_id, tkb)` (`pyrite/services/kb_service.py:757`) queries the SQLite index (`pyrite/storage/crud.py:22`), which in Pyrite is a *derived* cache — the markdown files are the source of truth. An entry that exists on disk but has not been `index sync`'d yet is invisible to it, so this rejects a link whose target is genuinely there. I confirmed it with a test on your branch: write two entries through `KBRepository`, do not sync, and `add_link` raises `EntryNotFoundError` for a target whose file I can `load()` one line earlier. The same test passes on #116, which validates with `KBRepository.load`.

This turns a silent-corruption bug into a false-rejection bug on the common `create` → `link` path, which is a smaller bug but still a regression for anyone whose index lags. Switching the lookup to the repository fixes it and keeps everything else you wrote intact:

```python
target_kb_config = self.config.get_kb(tkb)
if not target_kb_config:
    raise KBNotFoundError(f"KB not found: {tkb}")
target_repo = repo if tkb == source_kb else KBRepository(target_kb_config)
resolved = target_repo.load(target_id) is not None
```

**2. An unregistered target KB reports the wrong error.** With the index lookup, `add_link(..., target_kb="no-such-kb")` raises `EntryNotFoundError: Target entry not found: whatever in no-such-kb` — the KB does not exist at all, and the message blames the entry. The snippet above fixes this too by checking the KB config first (`KBNotFoundError`). Worth a test.

**3. The duplicate-link early return reports `resolved: True` unconditionally.** At `kb_service.py:755`, when the link already exists you return `{"resolved": True}` without checking the target. I verified: an entry whose frontmatter already carries a link to a nonexistent `ghost` returns `{"resolved": True}` on a repeat call. Since #97 is the mechanism behind #64's `title: null` dangling outlinks, the repeat call is exactly when a caller might hope to learn the link is stale. Either check the target on that path too, or leave it and say so in a comment — but the current value is a claim the code has not verified.

**4. Drop the `.gitignore` hunk.** It appends, under `# Auto-added by github-engineer`, patterns for `*.obj`, `*.so`, `*.whl`, `nem_stats.c`, `*.exp`, `*.lib`. `nem_stats.c` is from an unrelated project — this looks like tooling on your machine adding it without asking. It is also **the only file that conflicts with `dev`**; the three source files and the test file all merge cleanly. Removing it makes the rebase almost free.

**5. `ruff format` fails on your new code.** CI runs `ruff format --check` and it would fail here:

```
-            raise EntryNotFoundError(
-                f"Target entry not found: {target_id} in {tkb}"
-            )
+            raise EntryNotFoundError(f"Target entry not found: {target_id} in {tkb}")
```

Run `ruff format pyrite/` and re-commit. (`ruff check` passes; it is only the formatter.)

**6. Rebase onto `dev`.** The PR is currently `DIRTY`. After (4) the only conflicting file is gone, so `git fetch origin && git rebase origin/dev` should be clean. I verified this by merging `origin/dev` into your head locally: `.gitignore` is the only `UU`, every other path auto-merges.

**7. Tidy the commit message while you are rebasing.** The subject is the issue title truncated mid-word — `fix: resolve #97 - kb_link validates the source entry but not the target — sile`. Better as a short statement of what the commit does (`fix: validate kb_link target, with allow_dangling for forward references`), with `Fixes #97` on its own line in the body so the issue closes on merge.

**8. Add a CHANGELOG entry.** Every user-visible change gets a line under `## [Unreleased]` → `### Fixed`, with the issue number. #116 has one you could model it on. Also worth mentioning in it that `pyrite link` on the CLI is now strict — the CLI calls the same `add_link` and has no `--allow-dangling` flag, so CLI users get the new validation with no escape hatch. Adding the flag to `pyrite/cli/entry_commands.py:499` would be a nice follow-up, but it does not have to be in this PR.

### Note on CI

**CI has never run on this PR.** The workflow run for `565a6f5` is `conclusion: action_required` — GitHub holds workflow runs for first-time contributors until a maintainer approves them. Nothing here has been machine-verified; the 4224-passing number above is from my own local run on your head. Only @markramm can release that gate.

### Conflict with PR #69

No textual conflict: #69 touches `pyrite/models/base.py`, `pyrite/storage/repository.py`, `pyrite/storage/index.py` and the CLI, and does not touch `kb_service.py`. It rewrites how frontmatter is emitted on save, which is downstream of your change, so this does not need to wait for it. Worth re-running the link assertions once #69 lands, but that is our job, not yours.

### Recommendation to the maintainer

**Merge #108 after the changes above — it is the right PR of the two, but not in its current state.** #108 is the only one that implements #97's full acceptance criteria (`allow_dangling` + `resolved`, which #116 omits entirely), and its tests are broader and verified red. But #116 gets the one thing #108 gets wrong: validation against `KBRepository.load` (the files) instead of `self.db.get_entry` (the derived index), plus the `KBNotFoundError` check for an unregistered target KB. **Credit to @YaoSong808 for the disk-based lookup, the KB-config check and the CHANGELOG entry — items 1, 2 and 8 above are their code and should be folded into #108 as such.** #116 also has a test asserting the source entry's stored links are unchanged after a rejected link, which is a good check #108 lacks and is worth lifting.

If the maintainer would rather not wait on a round trip: #116 is already green in CI, `DIRTY` only on `CHANGELOG.md`, and is a smaller correct-as-far-as-it-goes fix — merging it and opening a follow-up for `allow_dangling`/`resolved` is a defensible alternative, but it leaves two of the three acceptance criteria on #97 unmet, and the design work in this PR is the part worth keeping.

Either way, @fathirramadhan-web, thanks — this is careful work, and the escape-hatch design is better than what we asked for in the ticket. Happy to re-review as soon as you push.

fathirramadhan-web:
and

fathirramadhan-web:
and

markramm:
Thanks — this is the design #97 asked for, and it lands second on top of #116, whose disk lookup is the source of truth. I rebased onto #116's cleaned head (dropping the unrelated `.gitignore` hunk) and added one fixup commit credited to you and @YaoSong808: the target lookup is now index-first with the repository as fallback, so an entry created but not yet synced still validates; `resolved` on the duplicate path is computed rather than assumed; CHANGELOG line; `ruff format`. Your `allow_dangling`/`resolved` design and tests are unchanged.

One correction to something I said in the earlier review, because this time I measured it rather than assuming it. I expected the index-first lookup to make a *rejected* link cheap. It does not:

```
                    with index-first    disk only
accepted link           2.85 ms/call    2.90 ms/call
rejected link         163.18 ms/call  161.58 ms/call
```

`KBRepository.find_file` already resolves a *present* id by filename in 0.027 ms — quicker than the index row read at 0.087 ms — so there was nothing to win on a hit. The ~550 ms I quoted earlier is the *miss* path: when the filename lookup fails, `find_file` falls through to parsing the frontmatter of every file in the KB (146 ms at 400 entries), and a rejected link is by definition a miss, where consulting disk is required for correctness anyway. So the cost is real, but it lives in `find_file` rather than in `add_link`, and it is filed separately as #135 instead of being fixed here. I kept the index-first ordering because it is the right shape and it is what lets #135 be fixed without touching this code again — not because it is a speedup today.

CI is approved on this PR.

markramm:
Thanks for the contribution!

PR #116, opened 2026-09-18
fix: reject dangling kb_link targets

### LINKED ISSUE #97: kb_link validates the source entry but not the target — silently creates dangling links (mechanism behind #64)

YaoSong808:
`kb_link` now refuses to write a link unless both endpoints exist, preventing misspelled or deleted target IDs from becoming silent dangling outlinks.

1. Validate the target against the target KB's repository before duplicate detection or any source write. This uses the same disk source of truth as source validation and supports cross-KB links.
2. Return `LINK_FAILED` with `retryable: false` when either endpoint is missing, so agents do not retry a deterministic error.
3. Add regression coverage for missing sources and missing cross-KB targets, including proof that a rejected target is not persisted.
4. Document the user-visible behavior in the changelog.

Fixes #97

### Evidence

- `87 passed` across the MCP and related link suites.
- The repository's full pre-push suite passed on the current `dev` (`4221 passed, 69 skipped` in total). This host's sandbox DNS mapping for `example.com` was isolated for the three mocked web-clipper tests; those three and the remaining 4218 tests all passed.
- `ruff check pyrite/ tests/` and `ruff format --check pyrite/ tests/` pass.
- Both new regression tests were observed failing before the implementation change.

### Scope and parallel work

This intentionally uses the strict-validation alternative in the issue and does not add an `allow_dangling` argument or change the successful response shape. PR #108 appeared while this branch was being prepared; unlike that parallel implementation, this version validates the repository source of truth before duplicate detection, so an already-dangling duplicate cannot return success and a stale index cannot reject an existing target.

🤖 Generated with Codex


markramm:
## Conductor partial review (tick 5) — handing off to a dedicated reviewer

Context: a dedicated review agent is taking this PR over. Posting what I verified so it is not lost, and releasing my `in-review` claim. **No merge decision here — outside contributions are the maintainer's call.**

**CI has never run on this PR.** The workflow run for `e042f80` is `conclusion: action_required` — GitHub is holding it pending maintainer approval for a first-time contributor. So "full suite green" is the author's report only; nothing has been machine-verified. Only the maintainer can approve the run.

**The disk-vs-index claim in the PR description is correct**, and it is the substantive difference from #108. This PR validates with `KBRepository(target_kb_config).load(target_id)` — the filesystem, the source of truth. #108 validates with `self.db.get_entry(target_id, tkb)`, which is the SQLite index (`pyrite/storage/crud.py:22`). An entry created on disk but not yet `index sync`'d exists but is not in the index, so **#108 will reject a link to a target that really is there**. That is a false negative this PR does not have, and it matters because `pyrite create` followed immediately by `pyrite link` is the common agent path.

Also correct here: the KB-not-found case raises `KBNotFoundError` rather than silently defaulting, and validation happens **before** any write to the source entry, so a rejected link cannot leave a half-written file. The regression test asserts that directly by reloading the source and checking no link was persisted — the right shape of assertion.

Points for the dedicated reviewer to weigh, not defects I confirmed:
- `mcp_server.py` catches `EntryNotFoundError` before `PyriteError`; worth confirming the former subclasses the latter, or the ordering is load-bearing in a fragile way.
- `KBNotFoundError` propagates out of `add_link` and is not caught by the new `except EntryNotFoundError`, so it falls to `except PyriteError` and returns `retryable: true`. A missing KB is not retryable either.
- This PR offers no forward-reference escape hatch; #108 adds `allow_dangling`. Whether forward references are a real use case in this KB model is a design question for the maintainer.
- Conflict with #69: both touch the save path only indirectly — #69 rewrites `Entry.to_markdown`/frontmatter emission, this touches `kb_service.add_link` before the save. I saw **no textual conflict** between them, but #69 changes what `add_link`'s eventual `save_entry` writes, so this PR's test assertions about stored links should be re-run after #69 lands.

markramm:
Thanks for this — the maintainer has read both PRs for #97 and this is the approach we want to build on.

The reason is the one you identified yourself: validating against `KBRepository.load` reads the filesystem, the source of truth, while validating against the index (`self.db.get_entry`) misses an entry that exists on disk but has not been `index sync`'d yet. Since `pyrite create` followed immediately by `pyrite link` is the common agent path, index-based validation would reject links to targets that really are there. Reading disk is correct.

Two other things this gets right and worth naming: validation happens **before** any write to the source entry, so a rejected link cannot leave a half-written file; and the regression test asserts that by reloading the source and checking no link was persisted, rather than only asserting the error. That is the right shape of assertion.

### Before it lands

**CI has not run yet.** The workflow run for `e042f80` is `action_required` — GitHub holds runs on first-time contributors until a maintainer approves. The maintainer is approving it; nothing is needed from you. Until it goes green, the suite result is your report rather than a machine-verified one, which is the only reason this is not merged already.

Three points we would like your view on. None is a blocker:

1. **`KBNotFoundError` is not caught by the new handler.** It propagates out of `add_link`, past `except EntryNotFoundError`, into `except PyriteError`, and comes back as `retryable: true`. A missing KB is no more retryable than a missing entry — it probably wants the same non-retryable `LINK_FAILED` treatment.

2. **Exception ordering.** `except EntryNotFoundError` sits before `except PyriteError`. If the former subclasses the latter the ordering is load-bearing and correct, but worth a comment saying so, since reordering them later would silently change the `retryable` flag.

3. **No forward-reference escape hatch.** The competing PR (#108) adds `allow_dangling` plus a `resolved` flag in the response, so a caller can deliberately link to an entry it is about to create and can tell resolved links from unresolved ones. That is a genuinely good idea. Whether it belongs here or in a follow-up is open — if you would like to add it, we would take it; if you would rather keep this PR tight, we will file it separately and credit #108.

Happy to merge once CI is green and you have had a chance to respond to (1).

markramm:
Thanks for this, @YaoSong808 — and thanks in particular for reading #97 carefully enough to choose the strict-validation option deliberately and say so in the PR body. Reviewing a contribution is much easier when the scoping decision is written down rather than left to be inferred.

I'm a reviewer agent working for the maintainer; the merge decision is his. Here is what was checked and what was found.

## What was checked

- Every hunk of `git diff origin/dev...HEAD` (4 files, +61/−3), read twice — once by me, once by a second reviewer given the diff and no other context.
- Full suite in a clean worktree at your head `e042f80`: **4222 passed, 68 skipped**, run twice with the same result.
- `ruff check pyrite/ tests/` and `ruff format --check` — clean (397 files already formatted).
- **Verified both tests are genuinely red without the fix.** Reverting `kb_service.py` and `mcp_server.py` to `dev`: `test_kb_link_target_not_found` → `KeyError: 'error_code'`, `test_kb_link_not_found` → `assert True is False`. Your "observed failing before the implementation change" claim holds up.
- Traced every caller of the changed method, plus the link-writing paths that don't call it.
- Measured the cost of the added lookup against this repo's own 773-entry KB.
- Conflict check against the open branches, including PR #69, and a full read of the competing PR #108.

## What is good

- **You fixed more than the title claims, and correctly.** `KBService.add_link` is shared by three write paths, not just MCP: `pyrite link`, `pyrite create --link` (`entry_commands.py:294`) and `kb_link`. Putting the check in the service fixes all three at once. Both CLI callers already catch `PyriteError`, so they degrade sensibly with no change needed.
- **Validating the target KB as well as the target entry** is a real gap in the competing PR that you closed.
- **`target_repo = repo if tkb == source_kb else KBRepository(...)`** correctly reuses the source repository for the same-KB case instead of building a second one.
- **Linking *into* a read-only KB still works.** You only `load()` the target, never write it, so the read-only guard stays on the source where it belongs. Easy to get wrong; you didn't.
- **The persistence assertion is the best line in the tests** — asserting the rejected target is absent from the *reloaded* entry's links tests the actual bug from #64, not just the error shape. That is the one test here that genuinely guards the service change.
- The changelog entry is accurate as far as it goes and is written for a user.

## Requested changes

**1. The existence check is inserted above the duplicate check, which breaks idempotent re-linking.** (`kb_service.py:748` vs the duplicate loop at `:752`)

Link A→B, delete B, then re-issue the identical link. Previously the duplicate loop matched and returned silently; now it raises before reaching that loop. Reproduced through the MCP dispatcher:

```
1st link : {'linked': True, 'source_id': 'idem-source', 'target_id': 'idem-target', ...}
delete   : {'deleted': True, 'entry_id': 'idem-target'}
re-link  : {'error': 'Entry not found: idem-target', 'error_code': 'LINK_FAILED', 'retryable': False}
```

Anything that replays a link set for idempotency — a re-run migration, a re-driven bulk script, an agent retrying a batch — now fails on links already correctly recorded in the source's frontmatter.

I want to flag honestly that this is in tension with your PR body's argument for the current order, and your argument is not wrong: checking first is what stops an *already-dangling* duplicate from returning success. Both properties are desirable. The way to get both is to keep the duplicate loop first but have it fall through to validation when the existing link is itself unresolved, rather than ordering one ahead of the other wholesale. Worth a test either way, because the two behaviours pull in opposite directions and the next person will need to know which was intended.

**2. A nonexistent `target_kb` still returns `retryable: true`** — the same wrongness #97 called out. Your new `KBNotFoundError` at `kb_service.py:747` is not an `EntryNotFoundError` subclass, so it falls past your new handler into `except PyriteError` → `retryable=True`. Reproduced:

```
kb_link(source_id=<exists>, source_kb="test-events",
        target_id="anything", target_kb="no-such-kb")
→ {'error': 'KB not found: no-such-kb', 'error_code': 'LINK_FAILED', 'retryable': True}
```

An unregistered KB does not become registered on retry. `except (EntryNotFoundError, KBNotFoundError)` in `_kb_link` fixes it. Note this is the precise anti-pattern `tests/test_mcp_server.py:165` already exists to prevent ("sending agents into pointless retry loops on a deterministic error") — so this PR makes one sibling case non-retryable while introducing another that is.

**3. The target lookup is very expensive, and the cost lands on the failure path.** (`kb_service.py:749`)

`KBRepository.load()` → `find_file()` falls through to reading and YAML-parsing **every** `.md` file in the KB before returning `None` (`storage/repository.py`, the `rglob("*.md")` fallback). So a rejected link pays a full-KB scan and parse, and so does a *successful* link to any entry whose id ≠ its filename — every ADR, where the file is `0025-release-workflow.md` and the id is `adr-0025`.

Measured against this repo's `kb/` (773 entries), on an idle machine, median of 7 runs:

```
id == filename (e.g. "roadmap")     0.4 ms     baseline
link rejected (missing target)    551   ms   ~1,400x baseline
id != filename (e.g. "adr-0025")  658   ms   ~1,650x baseline
```

So **rejecting a link costs about half a second**, and so does creating a perfectly valid link to any ADR. There is no cache, so N bad links in a batch cost N full scans, and it grows linearly with KB size — a 5,000-entry KB would be several seconds per call. The second reviewer measured this independently and isolated where it goes: `rglob` itself is 0.004 s; the frontmatter read-and-parse of all 773 files is essentially the whole cost. (Their run showed ~1.5 s on a loaded machine — same shape, and the ratio is the stable number either way.)

`self.db.get_entry(target_id, tkb)` is an O(1) indexed lookup, and this same service already uses it for exactly this purpose one method away in `rename` (`kb_service.py:700`). Your PR body is right that a bare index check can wrongly reject an entry the index hasn't caught up with — so use it as the fast path and keep your scan as the fallback:

```python
if not self.db.get_entry(target_id, tkb):          # O(1)
    if not target_repo.load(target_id):            # only when the index says no
        raise EntryNotFoundError(f"Entry not found: {target_id}")
```

This keeps your correctness property intact and pays the expensive scan only on the path that ends in an error anyway. Worth a test that a target present on disk but missing from the index is still accepted, since that is the case you're protecting.

**4. The changelog promises an invariant the code does not enforce.** `pyrite links bulk-create` (`cli/link_commands.py:277`) bypasses `KBService.add_link` entirely — it loads and checks the *source*, then calls `entry.add_link()` and `repo.save()` directly with no target check. Verified:

```
pyrite links bulk-create  (spec targeting a nonexistent entry)
exit: 0
Bulk create: 1 created, 0 skipped, 0 failed
entry-a links: [('ghost-does-not-exist', 'test-notes')]
```

A reader of "a misspelled or deleted target cannot create a dangling outlink" will believe that holds everywhere; on the bulk path — the one most likely to carry a typo — it does not. Either route bulk-create through the service or narrow the changelog sentence to the paths actually covered. (The extension hooks at `extensions/cascade/.../hooks.py:141` and `extensions/journalism-investigation/.../hooks.py:49` call `entry.add_link` directly too. If the intended invariant is "no dangling outlinks anywhere", the enforcement point is `Entry.add_link` or save time rather than one service method — but that is a larger design question, not something to ask of this PR.)

**5. Minor — `target_id=None` now raises `AttributeError`.** `find_file` calls `entry_id.startswith(...)`, and `AttributeError` is caught by neither `except PyriteError` in `_kb_link` nor `except (PyriteError, ValueError)` in `entry_commands.py:499`, so it escapes as an internal error. Previously `None` was silently written. Both are wrong; this swaps a silent corruption for a crash. A guard costs one line.

**6. Minor — the error code.** `mcp_server.py:120` has a `_DOMAIN_ERROR_CODES` table mapping `EntryNotFoundError` → `NOT_FOUND`; this handler hard-codes `LINK_FAILED`, so `kb_link` reports a missing entry differently from every other tool. #97's acceptance text does say `LINK_FAILED`, so I'd leave it — but a one-line comment noting the deliberate divergence would save the next reader the trip.

**7. Optional — no CLI-level test.** That the service fix also covers `pyrite link` and `create --link` is a genuine strength of your approach, but nothing pins it, and the two surfaces differ (`pyrite link` exits 1; `create --link` exits **0** with a yellow "Link failed" and creates the entry without the link). One test would stop a future refactor of `add_link`'s exception types from quietly changing the CLI contract.

One note on the test pair: `test_kb_link_not_found` passes with the service hunk reverted — only the MCP handler change makes it green. So the two tests read like two guards but are one; `test_kb_link_target_not_found` is the only one that fails without the service change. Not a defect, just worth knowing.

## Things the maintainer will want to know

- **CI ran and passed** on `e042f80` — no `action_required` approval gate was pending. The `pull_request` run completed `success` (test 3.12 and gate both green).
- **The PR is `CONFLICTING` with `dev`, but only in `CHANGELOG.md`** — `dev` added entries in the same `### Fixed` block. The code merges clean; it's a one-file rebase.
- **No conflict with PR #69** (`fix/cli-write-and-report-correctness`), despite that PR rewriting the entry save path. #69 touches `models/base.py`, `storage/repository.py`, `storage/index.py` and `cli/`; this touches `services/kb_service.py` and `server/mcp_server.py`. The only shared file is `CHANGELOG.md`, and a trial merge of the two heads is clean. **This does not need to wait for #69.**

## Recommendation to the maintainer

**Merge after changes 1–4** (5–7 are cheap and worth folding in), **and prefer this PR over #108.**

On **#116 vs #108** — same issue, and the choice is real:

- **#116 is the better-shaped base.** It validates the target KB as well as the entry, its tests assert persisted state rather than just the error shape, and its diff is small and free of unrelated noise. #108 returns early on a duplicate and never checks the target — and returns `{"resolved": True}` while doing so, which on an already-dangling link is an affirmatively wrong answer rather than merely a missed check. That is the #64 bug still reachable. #108 also carries an unrelated `.gitignore` hunk (`*.obj`, `*.so`, `nem_stats.c`, "Auto-added by github-engineer") that should not merge either way.
- **#108 is better on two specific points, and both should be taken from it:**
  1. **The index lookup** (`self.db.get_entry`) rather than a repository scan — request 3 above, and #108's single biggest advantage. #116's author was right that a bare index check can wrongly reject; the answer is index-first with a repository fallback, which is neither PR as written.
  2. **`allow_dangling` plus `resolved` in the response,** including the `tool_schemas.py` entry so agents can discover it. #97's acceptance criteria explicitly ask for this ("or, when the caller wants a forward reference, requires `allow_dangling: true` and reports `resolved: false` per #64"). #116 deliberately omits it and says so — a legitimate scoping call, but it means **#97 is not fully closed by this PR alone.** Either ask for it here or open the follow-up. It also gives request 1 an escape hatch: a caller replaying a link set can pass `allow_dangling` instead of failing.

So: take #116 as the base, ask for the index-first lookup, the `retryable` fix, the ordering fix and the changelog/bulk-create gap, and credit #108 for the index approach and the `allow_dangling`/`resolved` design — whether that lands here or in the follow-up that actually closes #97.

Thanks again, @YaoSong808. This is careful, well-scoped work, the target-KB validation and the persistence assertion are things the competing PR missed, and the PR body's reasoning about the index made the review genuinely easier even where I'm asking you to go the other way.

markramm:
Thanks — this is the right fix and it merges first of the two PRs for #97. I rebased onto `dev` (one CHANGELOG hunk) and added one fixup commit on top of yours, credited to you as co-author: the existence check now sits below the duplicate loop so re-issuing an existing link stays a no-op even if the target has since been deleted, and a missing target KB reports `retryable: false`. Your disk-based lookup is kept as the source of truth; `allow_dangling`/`resolved` and the index-first fast path land in #108 on top of this, with your check as its fallback. One deliberate trade-off, recorded in the fixup's commit message: with the duplicate check first, re-issuing a link whose target is already gone returns success rather than an error; the clean fix for that is the `resolved` flag, which is #108's scope. CI is running on the pushed commit.

markramm:
Rebased once more onto `dev` after Playwright package C landed (CHANGELOG hunk only; your commit unchanged, the maintainer's "update branch" merge commit replaced by the linear rebase the `dev` ruleset wants). Same two commits: yours, then the credited fixup. CI needs the maintainer's approval click on this head; once green it merges.

PR #126, opened 2026-09-18
fix: kb_batch_read no longer raises KeyError when fields omits kb_name

### LINKED ISSUE #57: kb_batch_read: raw KeyError 'kb_name' when fields omits kb_name; retryable is wrongly true
Source: read-tier hallway test, 2026-09-18, cold-start session over the MCP tools (54 KBs, 27,135 entries). Every error message quoted verbatim from that session.

```
kb_batch_read(entries=[{entry_id, kb_name} × 3], body_limit=5000,
              fields=["id","title","entry_type","tags","body","links"])
```
```json
{"error":"'kb_name'","error_code":"INTERNAL","retryable":true}
```

The message is the bare Python dict key. `retryable: true` is wrong — the identical call fails identically. Adding `"kb_name"` to `fields` fixes it, which is not discoverable from the error.

Fix: default `kb_name` (and `id`) into the projection, or fail with `VALIDATION` and `fields must include 'kb_name'`. Compare the `QUERY_SYNTAX` error, which is the model: states the constraint, gives the remediation, `retryable: false`.

makiaveli1:
## Why

`kb_batch_read` crashed with a raw `KeyError` when the caller's `fields` list omitted `kb_name` (or `id`):

```json
{"error":"'kb_name'","error_code":"INTERNAL","retryable":true}
```

The handler projects each entry down to the requested fields, then computes `found_ids` from `r["id"]`/`r["kb_name"]` — keys the projection had just dropped. `_dispatch_tool`'s catch-all turned the KeyError into `INTERNAL, retryable=True`, sending agents into retry loops on a call that fails identically every time. Same class of deterministic error the `QUERY_SYNTAX` contract already handles for search.

## Scope

- `pyrite/server/mcp_server.py` (`_kb_batch_read`): the projection now always includes the identity pair, `id` and `kb_name`, so `not_found` can be computed and the response stays well-formed whatever `fields` asks for.
- `pyrite/server/tool_schemas.py`: the `fields` description now says `id` and `kb_name` are always included.
- `tests/test_mcp_server.py`: `test_kb_batch_read_fields_without_kb_name_is_not_internal` — one real entry plus one missing ID, `fields: ["title", "entry_type"]`; asserts no `error_code`, the `not_found` list, and the identity pair in the projected entry.

I took the first option from the issue — default the identity pair into the projection — rather than failing the call with `VALIDATION`: the reported call itself is reasonable, and the response contract (`found`/`not_found`) is keyed by that pair.

## Evidence

- The new test fails on the unfixed code with exactly the reported shape: `{'error': "'id'", 'error_code': 'INTERNAL', 'retryable': True}`.
- Full suite (`pytest tests/ extensions/ -n auto`): **4221 passed, 69 skipped**.
- Full suite with CI's warning escalation (`-W error::DeprecationWarning:pyrite -W error::DeprecationWarning:tests`): **4221 passed, 69 skipped**.
- `ruff check` and `ruff format --check` clean on the touched files.

Fixes #57


markramm:
Thanks for this, and welcome. I'm the project's automated conductor; I review outside PRs so a maintainer can decide quickly. A maintainer makes the merge call, so treat this as a thorough read rather than a decision.

## What I checked

- Checked out `62d24b2` (matches the PR head) in a clean worktree.
- Read every hunk of `git diff origin/dev...HEAD` — all three files.
- Full suite: `pytest tests/ extensions/ -n auto` → **4222 passed, 68 skipped, 53s**. Green.
- Verified the new test is load-bearing: with `mcp_server.py` reverted to the merge base, `test_kb_batch_read_fields_without_kb_name_is_not_internal` **fails**; with the fix it passes. That check exists to catch tests that read as tested but only cover cases that already passed. Yours passes it.
- `ruff check` / `ruff format --check` clean; no absolute home paths.
- A second reviewer read the diff cold — no ticket, no PR body, no conversation — as we do for anything touching `pyrite/server/` or a public shape. I reproduced its findings myself before repeating them below.
- Checked conflicts against the other open PRs and against what has landed on `dev` since you branched.

## What is good, specifically

Three things here are better than a typical first patch, and I want them on the record:

1. **The diagnosis is exactly right, and it is the non-obvious part.** The handler projects each entry down to `fields`, *then* computes `found_ids` from `r["id"]` / `r["kb_name"]` — keys the projection just removed — and `_dispatch_tool`'s catch-all turns the `KeyError` into `INTERNAL, retryable=True`. You fixed it at the real defect rather than at the symptom.
2. **You chose between the issue's two options and said why.** Defaulting the identity pair in, rather than failing with `VALIDATION`, is the right call, and your reasoning is the right reasoning: the reported call is legitimate, and the `found`/`not_found` contract is keyed by that pair. Failing a reasonable request to protect an internal computation would be the API apologising for its own shape.
3. **You matched an existing precedent instead of inventing one, and you updated the schema in the same change.** `_kb_search`'s `QUERY_SYNTAX` handler already documents this class of mis-labelled deterministic error; your comment follows that form. And `tool_schemas.py` is what an agent actually reads before calling the tool — changing behaviour and leaving the description stale would have traded a crash for a quieter surprise. That step is the one most first patches skip.

I also confirmed the fix cannot hurt existing callers: `_project_fields` is a dict comprehension, so a caller who *did* request `id`/`kb_name` sees duplicates collapse. Only key ordering shifts, which is not a contract.

## Requested changes

**1. Fix the identical bug two lines above your patch (`mcp_server.py:535`).**

This is the one that matters. `ids = [(e["entry_id"], e["kb_name"]) for e in entries_spec]` raw-indexes caller-supplied dicts, immediately after two explicit `VALIDATION_FAILED` guards — so the handler already owns its input validation. I reproduced this on your branch, through the real dispatcher:

```
kb_batch_read(entries=[{"kb_name": "test-events"}])
  -> {"error": "'entry_id'", "error_code": "INTERNAL", "retryable": true}
kb_batch_read(entries=[{"entry_id": "x"}])
  -> {"error": "'kb_name'",  "error_code": "INTERNAL", "retryable": true}
```

That is the same user-visible failure #57 describes — bare key name as the message, `INTERNAL`, `retryable: true` telling an agent to retry a call that is deterministically wrong — still reachable through the same tool, from an equally plausible argument (an agent that built `entries` from a search result and dropped a key). A fix that leaves its own twin live in the same function is the thing that undermines confidence in the fix. A `VALIDATION_FAILED` alongside the existing two guards, naming the missing key, would close it.

**2. Add a `CHANGELOG.md` entry under `## [Unreleased]` → `### Fixed`.**

This repo keeps a line per user-visible change, and an MCP tool that stops returning `INTERNAL` on a valid call qualifies. Something like:

> - `kb_batch_read` no longer fails with a raw `KeyError` reported as `INTERNAL`/`retryable=true` when `fields` omits `id` or `kb_name`; the identity pair is always included in the projection. (#57)

**3. Narrow the schema sentence, or widen the fix.**

`tool_schemas.py:270` now promises "`id` and `kb_name` are always included", but that is true only of `kb_batch_read`. `kb_get` (`:372`), `kb_search` (`:341`), `kb_list_entries` (`:582`) and `kb_recent` (`:622`) all still project bare. None of them can hit *this* crash — none reads the identity back out afterwards — so they are not bugs in the #57 sense, but they do return rows an agent cannot act on (a list of `{"title": ...}` with no id is not useful for a follow-up fetch). Either scope the sentence to this tool, or — if you'd rather do it properly — give `_project_fields` a `keep=("id","kb_name")` argument and apply it at all five call sites, which makes the sentence true tool-wide. Either is fine; the current state promises more than it delivers.

Optional, no action needed from you: the branch is behind `dev` (merge base `3fe192c`), but nothing that landed since touches these three files, so the rebase is trivial and a maintainer can do it with `gh pr update-branch --rebase`.

## Not yours to fix — filed for the maintainer

Two things the cold read surfaced that pre-date this PR. I'm recording them so they don't get lost when #57 closes, **not** asking you to take them on:

- **`pyrite/server/endpoints/entries.py:356,363,365` — REST `POST /entries/batch` is the same contract and now diverges from MCP.** It has the same unguarded `e["entry_id"]` (a 500 there, not a structured error), and its projection adds no identity pair while `found_ids` uses `.get()`. I reproduced the consequence: with `fields=["title"]` it returns `found: 2` *and* lists both entries in `not_found` — a self-contradicting response. Nothing tests the REST `fields` path.
- **`mcp_server.py:546` — `body_limit` is ignored when `fields` includes `body`.** `_chunk_body` is only in the `else` branch, so a projection containing `body` bypasses chunking *and* the `MAX_BODY_CHUNK = 50_000` ceiling, with no `body_truncated` flag for the caller to notice.

## Against the issue's acceptance

#57 raised three things:

- **Raw `KeyError` when `fields` omits `kb_name`** — fixed and tested. ✅
- **`retryable: true` on a deterministic error** — fixed for this path, by removing the crash entirely, which is the better outcome: there is no longer an error to mis-label. The catch-all at `mcp_server.py:1725` still defaults to `retryable=True` for any unexpected exception; that is a broader design question and does not belong in this PR.
- **The same leak in `kb_get`'s `fields` / `update` paths** — see requested change 3 above.

## Conflicts with other open work

None. PRs #116 and #108 touch `mcp_server.py` and `tests/test_mcp_server.py`, but in `_kb_link` (~lines 954–976) and tests from line 912; your change is at line 544 and test line 177. #108 also touches `tool_schemas.py` at line 834 vs your 267. Whichever merges first, the others rebase cleanly. Nothing here needs to wait.

## CI

So the state isn't misread: the workflow run for `62d24b2` is **`action_required`**. GitHub holds runs from first-time contributors until a maintainer approves them, so **the checks have not spoken on this branch yet**. That is repo policy, not a comment on your PR. The suite is green in my worktree — evidence, but not the gate.

## Recommendation to the maintainer

**Merge after these changes** — specifically item 1 (the sibling `KeyError` at `mcp_server.py:535`), plus the changelog line and the schema-sentence scoping. The fix is correct, minimally scoped, follows the file's existing error-contract precedent, and is pinned by a test that genuinely fails on revert. It is not a merge-as-is only because it leaves the identical failure reachable four lines from the patch, which I reproduced in two calls. Worth opening follow-up tickets for the REST `/entries/batch` divergence and the `body_limit` bypass.

Thanks again — the diagnosis, the stated trade-off and the evidence in your PR description did a lot of the reviewer's work, and that is genuinely appreciated.

markramm:
Merging this as-is — thank you, the diagnosis and the fix are right and the test proves it. The three changes the review asked for are now a follow-up issue rather than another round on this PR: #137 (the unguarded `entry_id` two lines above your patch, a CHANGELOG line, and narrowing or widening the "always included" schema sentence). It is labelled `good first issue` and yours if you want it; either way, welcome aboard.

Issue #49, opened 2026-09-18
Root layout overwrites every page title with the brand name — no route's <svelte:head><title> survives


markramm:
Found while building the deterministic Playwright world (Package A of the `playwright-e2e-suite-non-deterministic-failures-*` ticket).

## What happens

`web/src/routes/+layout.svelte` sets `document.title` from the branding store:

```js
// Push branding into the DOM as it loads:
//   --brand-primary      — accent color referenced by chrome components
//   document.title       — replace app.html's static "Pyrite"
$effect(() => {
    if (typeof document === 'undefined') return;
    document.documentElement.style.setProperty('--brand-primary', brandStore.primary_color);
    if (brandStore.loaded) {
        document.title = brandStore.name;
    }
});
```

Every route also declares its own title, e.g. `web/src/routes/search/+page.svelte`:

```svelte
<svelte:head><title>Search — Pyrite</title></svelte:head>
```

The layout effect runs once `brandStore.loaded` flips and clobbers whatever the page set. SSR is off (`+layout.ts`: `export const ssr = false`), so the page title is only ever applied client-side — after which the layout overwrites it.

## Evidence

Probe navigating to every top-level route with a 1.5 s settle, reading `page.title()`:

```
ROUTE /            -> "Pyrite"
ROUTE /entries     -> "Pyrite"     (page declares "Entries — Pyrite")
ROUTE /search      -> "Pyrite"     (page declares "Search — Pyrite")
ROUTE /settings    -> "Pyrite"     (page declares "Settings — Pyrite")
ROUTE /qa          -> "Pyrite"     (page declares "QA Dashboard — Pyrite")
ROUTE /daily       -> "Pyrite"     (page declares "Daily Notes — Pyrite")
ROUTE /timeline    -> "Pyrite"
ROUTE /graph       -> "Pyrite"
ROUTE /collections -> "Pyrite"
ROUTE /login       -> "Pyrite"
```

No route wins. This is a user-visible bug, not only a test one: every browser tab, every bookmark and every window-switcher entry reads "Pyrite" regardless of where you are in the app.

It is also a source of moving e2e failures, because it is a race: `settings.spec.ts:4` and `search.spec.ts:4` failed their `toHaveTitle` in all five of five consecutive runs, while `qa.spec.ts:4` and `daily.spec.ts:4` — the same assertion, same cause — failed in four of five. Whether the page title or the branding effect lands last depends on when `/config/branding` returns.

## Suggested fix

[...]

PR #138, opened 2026-09-18
test(e2e): Playwright package A.1 — per-worktree ports and data dir (#118)

### LINKED ISSUE #118: e2e Playwright suite: concurrent worktrees collide on hardcoded ports 8088/5173, causing cross-worktree data bleed and mid-run backend drops
## What's wrong

`web/playwright.config.ts` hardcodes the e2e backend to port 8088 and the Vite
dev server to port 5173, with `reuseExistingServer: false` (deliberately, per
the config's own comment -- reusing a server on 8088 might be someone's real
Pyrite instance). When two or more worktrees run their own `playwright test`
concurrently -- exactly what the conductor's parallel-package dispatch does --
they collide on these two ports, with two distinct failure modes observed
directly while working Package D
(`kb/backlog/playwright-package-d-entry-crud-and-entry-features-specs.md`):

1. **Vite silently falls back to a different port.** When 5173 is taken, Vite
   logs `Port 5173 is in use, trying another one...` and binds 5174 (or
   higher) instead -- but Playwright's `webServer.url` health check
   (`http://localhost:5173`) and `use.baseURL` are both hardcoded to 5173.
   Whichever OTHER worktree's server is actually listening on 5173 answers the
   health check, so Playwright proceeds thinking its own server is up, and the
   browser then talks to a stranger's backend/frontend (wrong seeded data,
   wrong `.e2e-data`) for the run. This produced `element(s) not found` /
   `strict mode violation` failures on entries that exist in this worktree's
   own seed but not in whichever sibling worktree's world actually answered.

2. **The uvicorn backend on 8088 becomes unreachable mid-run** even when it
   bound successfully at start (confirmed with a single-worker, `--trace=on`
   run: Vite's proxy logged literal `ECONNREFUSED 127.0.0.1:8088` for every
   API call from partway through the run onward, while `ps` showed this
   worktree's own uvicorn process gone and its own Vite process still
   listening). System load during the observed run was 22-33 on a 10-core
   machine (several sibling worktrees' own webServer processes and Playwright
   worker pools running at once). Root cause is most likely resource
   starvation (the backend's event loop stalling long enough to fail its own
   health check, or being reaped under load) rather than an explicit kill by
   another process, but this wasn't confirmed via a jetsam/OOM log entry.

Both failure modes are indistinguishable from real app or spec bugs from
inside a single worktree's test output -- they show up as flaky
`ERR_CONNECTION_REFUSED`, strict-mode violations against data that "shouldn't"
be there, or plain timeouts. Package D worked around this by polling
`lsof -i :8088`/`:5173` before each run and discarding any run whose log
contained `trying another one`, `already used`, `ECONNREFUSED`, or `was not
able to start`, but that's a workaround at the calling shell, not a fix, and
it does nothing for CI (which presumably runs one worktree at a time, so may
not hit this at all -- but any future move toward running e2e suites
concurrently, or a developer running two worktrees locally, will).

## Suggested fix (not for me to make -- `playwright.config.ts` is Package
D's explicit read-only boundary)

Something in the direction of: derive the port (and `.e2e-data` directory
name) from the worktree path or a `PLAYWRIGHT_E2E_PORT` env var set by
`scripts/new-worktree.sh`, so concurrent worktrees never share a port; or add
a preflight check that fails fast (rather than silently degrading) when the
target port is already bound by a process outside this worktree.

## Where this was found

Package D, rewriting `web/e2e/entry-crud.spec.ts` and
`web/e2e/entry-features.spec.ts` against the seeded world, while Packages
B/C/E were running concurrently in sibling worktrees per the conductor's
tick-3 dispatch.

markramm:
## Theme: Playwright package A.1 — per-worktree ports and data dir (#118)

Closes #118. Milestone 0.24.2. `heavy: yes` (Playwright verification runs). Model: sonnet.

### Why

`web/playwright.config.ts` hardcodes the backend to 8088 and Vite to 5173 with `reuseExistingServer: false`. Two worktrees running `playwright test` at once collide: Vite silently falls back to 5174+ while the health check and `baseURL` stay on 5173, so a run talks to a *sibling worktree's* backend and world, and the uvicorn on 8088 drops mid-run under load. Packages C/D/E ran three at once on 2026-09-18 and their evidence was contaminated; Playwright has been serialized to one worktree at a time since. This package removes the reason.

### Acceptance

1. Ports and the data dir are derived per worktree: `PLAYWRIGHT_E2E_PORT` (backend) and `PLAYWRIGHT_E2E_VITE_PORT` when set, else a stable hash of the worktree path mapped into a high range (e.g. 20000–29999, backend and Vite distinct); `web/.e2e-data` becomes `web/.e2e-data-<port>` or equivalent so two worktrees never share a world. `scripts/new-worktree.sh` exports/records the chosen ports (a `.pyrite/e2e-ports` file or the config's env) so a human running the suite by hand gets the same ones.
2. A preflight in `global-setup.ts` fails fast — a clear error naming the port and the owning process (`lsof -i :<port>`) — when the target port is already bound by a process outside this worktree. Vite's silent fallback is defeated: `strictPort: true` on the Vite webServer command (or the config's equivalent) so a taken port is an error, never a different port.
3. The `baseURL`, both `webServer.url` health checks, and any `/api` proxy target read the same derived ports; nothing in `web/e2e/*.spec.ts` needs to change.
4. Evidence: two worktrees (this one and a throwaway second `scripts/new-worktree.sh` from `origin/dev` with this branch cherry-picked) run `npm run test:e2e` **concurrently**; each passes `seed.spec.ts` against its own world, `lsof` shows two distinct port pairs, and neither run's log contains `trying another one`, `already used`, `ECONNREFUSED` or `was not able to start`. Then five sequential runs in this worktree: the same failing set every time (the known eleven from package A, minus whatever B–E fixed), and `~/.pyrite/index.db` mtime unchanged.
5. The config comment about never reusing a server on 8088 stays true in spirit (never reuse a server this worktree did not start). CHANGELOG line. The Playwright ticket's "Package A.1" section appended with the port scheme and the evidence.

### Touches

Existing: `web/playwright.config.ts`, `web/e2e/global-setup.ts`, `web/e2e/fixtures.ts` (only if the data-dir constant lives there), `scripts/new-worktree.sh`, `.gitignore` (the new data-dir pattern), `CHANGELOG.md`, `kb/backlog/playwright-e2e-suite-non-deterministic-failures-likely-shared-state-auth-config-gap.md`.
New: none expected.
Out of scope: rewriting any spec file (F/G), `ci.yml` (H), the server's own port handling.

---

## Worker report (pyrite-worker, sonnet — appended by the host at 10:03Z; absorb at the next tick)

**Pushed:** `4f2585b == origin/feature/playwright-a1-ports`
**Commits:** 33a0a89 feat: derive Playwright e2e ports per worktree, fixing collisions (#118); 4f2585b kb: close A.1, record evidence, changelog. (Claim commit b374739 pre-existing.)
**Closes:** #118.

**What changed.** New `web/e2e/ports.ts` derives all four e2e ports (base backend, base Vite, auth backend, auth Vite) from a SHA-256 of the worktree's path, each in its own 2,500-wide band of 20000–29999 so they cannot collide with each other; `PLAYWRIGHT_E2E_PORT`/`PLAYWRIGHT_E2E_VITE_PORT` override the base pair. Unit-tested in `ports.test.ts` (5 tests — the RED→GREEN; `vite.config.ts` `test.include` gained `e2e/*.test.ts`). `global-setup.ts` and `auth-setup.ts` compute `PORTS` at module scope and export `E2E_BACKEND_PORT`, `E2E_VITE_PORT`, `E2E_BACKEND_URL`; data dirs are `web/.e2e-data-<port>` / `web/.e2e-auth-data-<port>`; both gained `preflightPort()` — an `lsof -i :<port>` check that throws naming the port and its owner before any world builds. `vite.config.ts` gained `strictPort: true` (only the auth world had it) and a proxy target read from `PLAYWRIGHT_E2E_PORT` (default 8088, so `npm run dev` is unaffected); `playwright.config.ts`'s webServer commands, health URLs and `baseURL` read the derived ports, and the Vite entry passes the port in `env` (Playwright merges `webServer.env` with `process.env` — verified in its source). `scripts/new-worktree.sh` writes `.pyrite/e2e-ports` via `web/e2e/print-ports.ts` (Node 24 runs `.ts` directly; relative imports need the `.ts` extension there).

Two gaps found and fixed in scope: five specs (`app`, `entry-crud`, `entry-features`, `graph`, `seed`) hardcoded `localhost:8088` for direct-to-backend calls and would have gone `ECONNREFUSED` — each now imports `E2E_BACKEND_URL` (one line per file; no spec rewritten). Playwright's default `testMatch` also matched `*.test.ts` and crashed loading vitest's `expect`; the config now has `testMatch: /.*\.spec\.ts/`.

**Evidence**
- **Concurrent, two worktrees** (this one + a throwaway from `origin/dev` with 33a0a89 cherry-picked): derived ports fully distinct (21764/24384/25926/29231 vs 20152/24536/26077/27864); an `lsof` + `ps` snapshot mid-run shows all eight ports LISTENing with connections only from their own worktree's Chromium/uvicorn/vite, and both seeds' `pyrite create` running concurrently from separate venvs. Both passed `seed.spec.ts` 10/10 against their own worlds. This worktree 128 passed / 0 failed; throwaway 126 passed / 2 failed (`qa.spec:4`, `search.spec:4` — #49's title race). **Zero matches** in either log for `trying another one` / `already used` / `ECONNREFUSED` / `was not able to start`.
- **Five sequential runs**, clean batch (an earlier batch under load 13–24 discarded as evidence): 126/126/126/125/126 passed; runs 6, 7, 8, 10 failed the identical set (`qa.spec:4`, `search.spec:4` — #49); run 9 additionally `app.spec:26` under residual Spotlight load, not reproduced.
- `~/.pyrite/index.db` mtime/size (`Sep 18 04:06:53`, 637804544) and `config.yaml` unchanged through the concurrent run and all 15 sequential runs.
- `npx vitest run`: 38 files, 393 passed (incl. the 5 new); `npm run check` 0 errors / 23 known warnings; `npm run build` ✓. Pre-push pytest skipped (no `.py` in the diff). `ruff check .`: 75 pre-existing findings, all in `scripts/scrape_appointee_details.py`, not in this footprint.

**KB:** the A.1 item `status: done` (hand-edited — dispatched before #69 landed), moved to `kb/backlog/done/` with a Result section; a "Package A.1 result" section appended to the parent Playwright ticket; `index sync`; CHANGELOG line under `### Fixed`.

**Changed.** New: `web/e2e/ports.ts`, `ports.test.ts`, `print-ports.ts`. Existing: `global-setup.ts`, `auth-setup.ts`, `playwright.config.ts`, `vite.config.ts`, `scripts/new-worktree.sh`, `.gitignore`, `CHANGELOG.md`, five specs (constant substitution only), the two backlog files.

**Unsure**
- The five spec edits are outside the literal "Touches" list; judged in scope because reverting just them reproduces `ECONNREFUSED` — the mechanism is not complete without them. A reviewer reading AC #3 literally may want them as a separate commit on the same branch.
- `strictPort: true` on the shared `vite.config.ts` also makes a taken 8088 a hard error for a human's `npm run dev` (previously silent fallback) — what the ticket asks for, but visible outside e2e.
- The auth pair in `.pyrite/e2e-ports` is informational only (no override read-back); small follow-up if wanted.

**Left:** nothing in scope. F/G rewrites, `ci.yml` (H), the server's own port handling untouched.


markramm:
## Conductor review — tick 8 (host), after the cold read

Reviewed the pushed head `4f2585b` in a fresh review worktree (vitest 393 passed, svelte-check clean, one full Playwright run on the derived ports 22440/23849/25031/28143), then dispatched `pyrite-reviewer` on the diff alone. The cold read found the design sound — bands disjoint by construction, the auth world mirrors the base, `testMatch` precedence verified, `new-worktree.sh` works with no `node_modules` — and six things worth fixing before merge, the first of which a reader of the diff cannot see: **the unit tests did not test the mechanism** (deleting the banding, or hardcoding the auth pair, left all five green because every assertion used one literal path).

The worker (same context) applied all six in `dcd0ffd`, verified here: `ports.test.ts` is now a property-style suite over 300+ generated paths (band membership, mutual distinctness, stability, auth-pair variance, overrides) and the reviewer's two mutations now fail 2/9 and 1/9; a non-numeric `PLAYWRIGHT_E2E_PORT` throws a one-line error instead of producing `NaN` (reproduced); `print-ports.ts` derives from `{}` and names the auth pair as derived, not settable; `preflightPort` uses `lsof -sTCP:LISTEN` and says it fails open without `lsof`; the bare `web/.e2e-data/` and `web/.e2e-auth-data/` ignores are back; CHANGELOG and README say `strictPort` now makes a taken 5173 a hard error for a human's `npm run dev`. Suite on the fix commit: 126 passed / 2 failed — the #49 pair, unchanged from every run since package B.

One flake seen during review, filed as #153 (entry-features specs fail instantly on the first run after the servers start) — not this branch's defect.

Disposition: ready, auto-merge. Closes #118. Playwright F/G can run in parallel from here.

PR #139, opened 2026-09-18
quality: pin the test runner — one pytest and xdist for every venv and every CI leg (#128)

### LINKED ISSUE #128: CI resolves pytest 9.1.1 on 3.12 but 9.0.2 on 3.13; the skew let classmethod fixtures pass the PR gate and break dev — pin the test runner
The CI matrix resolves a different pytest per interpreter: the 3.12 job and every worktree venv got **pytest 9.1.1**, the 3.13 job and the main checkout's venv got **9.0.2**. That skew is why #81's classmethod fixtures passed the PR gate (3.12) and broke `dev` on 3.13 — 9.1.1 tolerates a classmethod fixture, 9.0.2 does not collect it.

Two things worth doing:
- Pin pytest (and xdist) to one version in `pyproject.toml`'s dev extras so every interpreter and every worktree runs the same test runner; the classifier/uv cache already keys on `pyproject.toml`.
- Consider whether `@classmethod` fixtures should be caught by ruff/pytest config (`PT` rules) so a tolerant pytest cannot mask it.

Evidence: dev runs d3d223a and 83b1da5 (test (3.13) failure, test (3.12) success); fix in the PR that references this issue.

markramm:
## Theme: pin the test runner — one pytest and xdist for every venv and every CI leg (#128)

Closes #128 and the backlog item `pin-the-test-runner-one-pytest-and-xdist-version-for-every-venv-and-every-ci-leg`. `heavy: no`. Model: sonnet.

### Why

CI resolved pytest 9.1.1 on 3.12 and 9.0.2 on 3.13; every worktree venv got 9.1.1, the main checkout 9.0.2. #81's `@classmethod` fixtures passed the one-interpreter PR gate and broke `dev` twice on 3.13 (fixed in #129). The gate and the matrix must run the same runner.

### Acceptance

1. `pyproject.toml` dev extras pin `pytest`, `pytest-xdist` and any pytest plugins present (`pytest-asyncio`/`anyio`, `pytest-cov`, `pytest-timeout` — whatever `uv pip list` shows installed from the extras) to exact `==` versions — pick the versions currently resolved on 3.12 (9.1.1 for pytest) unless one of them lacks a 3.13 wheel, in which case pick the newest that all three interpreters install, and say which.
2. `tests/test_dev_process_config.py` asserts each pin is exact (`==`), so loosening is a visible diff.
3. ruff's `PT` (flake8-pytest-style) rule set enabled for `tests/` in `pyproject.toml`'s ruff config, with any resulting findings either fixed in the same PR (if mechanical and few) or the specific rules ignored with a one-line reason — a `@classmethod` fixture must be a lint error under any runner. Report the count.
4. CHANGELOG line; a note in `CONTRIBUTING.md`'s testing section that the runner is pinned and how to refresh a venv (`uv pip install -e ".[all,dev]"`).
5. Evidence: `.venv/bin/python -m pytest --version` in this worktree after reinstalling from the pinned extras; the full suite green; the PR gate's `test (3.12)` and a `gh workflow run ci.yml --ref <branch>`-style or post-merge matrix run showing identical pytest versions in all three job logs (or, if you cannot trigger the matrix pre-merge, say so and the reviewer will check post-merge).

### Touches

Existing: `pyproject.toml`, `tests/test_dev_process_config.py`, `CHANGELOG.md`, `CONTRIBUTING.md`, and `tests/**` only where a `PT` finding is fixed mechanically.
New: none.
Out of scope: anything in `pyrite/`; changing the CI matrix.

---

## Worker report (pyrite-worker, sonnet — appended by the host at 09:48Z; absorb at the next tick)

**Pushed:** `9f725ed == origin/quality/pin-test-runner`
**Commits:** 295fe77 pins pytest/pytest-cov/pytest-xdist exactly and enables `PT` for `tests/`; 68c56bb 14 mechanical PT fixes (fixture parens, useless yield, tuple parametrize); 9f725ed CHANGELOG/CONTRIBUTING, closes #128.
**Closes:** #128; backlog item `pin-the-test-runner…` (done, moved to `kb/backlog/done/`).

**Evidence**
- Pins `pytest==9.1.1`, `pytest-cov==7.1.0`, `pytest-xdist==3.8.0` in the `dev` extra — all pure-Python wheels with broad `requires_python` (checked on PyPI), so one version installs identically on 3.11/3.12/3.13. No other pytest plugin is installed from the extras (`anyio` is transitive via httpx/mcp; absent from pytest's plugin list).
- RED→GREEN: `TestPinnedTestRunner` in `tests/test_dev_process_config.py` failed against the unpinned file (`not pinned at all in dev extras: {'pytest-xdist','pytest','pytest-cov'}`), then 37 passed.
- Reinstalled the venv from the pins: `pytest --version` → 9.1.1; `uv pip list` confirms the other two.
- Full suite: **4240 passed, 68 skipped** (twice, ~51–64 s).
- `PT`: 51 findings before, all in `tests/` (none in `pyrite/`, `extensions/`, `scripts/`); 14 fixed mechanically by `ruff --fix` (PT001 ×7, PT022 ×4 — each checked to have no teardown, PT006 ×3); 37 ignored per-rule under `tests/*` with a one-line reason each (PT011, PT012, PT018, PT019 — all need per-call-site judgment, no autofix). `ruff check --select PT .` clean; `ruff format --check tests/` 212 files formatted. 75 pre-existing repo-wide lint findings unrelated (identical count with the change stashed).
- Footprint clean: nothing under `pyrite/`.

**Changed:** `pyproject.toml` (pins; `PT` in `select`, off outside `tests/` via `per-file-ignores`), `tests/test_dev_process_config.py`, nine test files (mechanical PT fixes only), `CHANGELOG.md`, `CONTRIBUTING.md` (pinned-runner paragraph + reinstall command), the backlog item.

**Unsure — one acceptance criterion not met, with evidence:** "a `@classmethod` fixture must become a lint error under any runner" cannot be satisfied with ruff or pytest: no PT rule (checked every rule in ruff 0.15.22's `flake8_pytest_style`, `--select ALL --preview`) and no pytest warning fires on `@classmethod` above `@pytest.fixture`. pytest's own source treats the *opposite* ordering (a class-scoped fixture as a plain instance method) as the deprecated pattern and tells you to use `@classmethod`. #81's fixtures were fixed forward by removing `@classmethod` because of the 3.13/9.0.2-specific bug that the exact pin now makes impossible. A static check would need a custom script + pre-commit hook (the repo's `scripts/check_import_cycles.py` pattern) — outside this theme's footprint; documented in the KB closure note.

**Left:** the criterion above (a considered "this mechanism does not exist"). Filed **#144**: `tests/test_api_tiers.py`'s five class-scoped fixtures are now on the *deprecated* instance-method pattern — a live `PytestRemovedIn10Warning` in every run; with the runner pinned, moving them back to `@classmethod` should be safe but must be verified against 9.1.1 specifically. No matrix run was triggered pre-merge; the reviewer checks the post-merge `dev` matrix for identical `pytest 9.1.1` in all three legs. The KB closure was hand-edited (`pyrite update` was #46 at dispatch time).


markramm:
## Conductor review — tick 8 (host)

Re-verified on the pushed head `9f725ed` in a fresh review worktree, not from the report:
- Reinstalled the venv from the pinned extras → `pytest 9.1.1`; full suite **4240 passed, 68 skipped** (79 s).
- `ruff check --select PT .` clean; `ruff format --check tests/` 212 files formatted.
- Reproduced the worker's finding about the unmet criterion: a `@classmethod` fixture under the pinned 9.1.1 is neither a lint finding nor a runtime failure — it passes. So the pin is what makes the #81 class of failure impossible, and no static check for the pattern exists in ruff or pytest (pytest treats the *opposite* ordering as the deprecated one). Criterion 3's premise was wrong; the worker's reasoning stands, and #144 tracks the now-deprecated instance-method fixtures in `test_api_tiers.py`.
- Diff read: pins are exact (`==`), `PT` is on globally and off outside `tests/` via per-file-ignores, four judgment-only rules ignored with reasons, the nine test-file edits are `ruff --fix` output only. No file under `pyrite/`.
- Left for post-merge: the `dev` matrix run must show `pytest 9.1.1` on all three legs — the conductor checks it on the merge's run.

Disposition: ready, auto-merge. No cold read (no server/storage/schema change; the Unsure names an unmet criterion with evidence, not a design decision).

PR #164, opened 2026-09-18
fix: stop entries inventing created_at/updated_at on save (#151)

### LINKED ISSUE #151: Entry.created_at / updated_at are read from frontmatter but never written back (silent drop on save)
## Problem

```python
```

pair of keys.

## Evidence

```yaml
---
id: created-as-date
type: backlog_item
kind: feature
priority: low
effort: S
status: planned
created_at: 2026-01-15
updated_at: 2026-01-15
---
```

it on the next save.

## Likely fix

had them don't grow them.

## Scope note

the fix).

zhongxiao-chang:
Closes #151

## Reproduction

On `dev` @ 36d4b59, in a KB whose entry file carries no timestamp keys:

```python
from pyrite.storage.repository import KBRepository
entry = repo.load_entry_from_file(path)   # frontmatter: id/type/title only
repo.save(entry)
# after: frontmatter now has updated_at  (key the file never had)
```

Related, same issue: a bare YAML date in frontmatter read back as the *load* time:

```python
parse_datetime(datetime.date(2026, 1, 15))
# before: datetime.now(tz=UTC)   <- the file's value was replaced by "now"
# after:  datetime(2026, 1, 15, 0, 0, tzinfo=UTC)
```

## Root cause

Two independent halves:

1. `created_at` / `updated_at` are read from frontmatter but never written back — `KBService.update_entry` writes the file from a snapshot taken before the in-memory stamp, so the key silently drops on save.
2. `parse_datetime()` only handled `datetime` / `str` and fell back to `_utcnow()` for everything else, so the `datetime.date` objects YAML produces for bare dates were mapped to "now".

## Fix

- `pyrite/models/base.py`
  - `parse_datetime()` gains a `date` branch: naive `date` → midnight UTC, matching `_utcnow()`'s UTC semantics; the tz contract is now stated in the docstring.
  - `_frontmatter_for_file` writes a default key back only when the source file carried it, or when the field was explicitly assigned (which clears it from `_absent_default_keys`). An untouched, absent key is never invented.
  - `touch_updated_at()` refreshes the stamp through `object.__setattr__`, so internal bookkeeping does not count as a user assignment and cannot resurrect a key the file never had.
- `pyrite/storage/repository.py`, `pyrite/services/kb_service.py`, `pyrite/cli/link_commands.py` — the four internal stamping sites (`KBRepository.save`, `KBService.update_entry`, `add_link` ×2, `sw link` bulk create) now call `touch_updated_at()` instead of assigning `updated_at` directly.
- Two tests added to `tests/test_update_preserves_frontmatter.py`: a repository-level file without timestamps must not gain them on save, and a file that carries them is still refreshed and written back.

## Verification

`tests/test_update_preserves_frontmatter.py`, `PYTHONPATH=extensions/software-kb/src`:

```
11 failed, 23 passed
```

The 11 failures are the same pre-existing failures as on the clean `dev` base (11 failed / 21 passed there) — **net new failures = 0**, plus 2 new passing tests. The remaining reds are environment reds in this clone (`pyrite` not `pip install -e .`), not related to this change.

`ruff check` on the five touched files: **All checks passed** (the change's four dead `datetime` imports were removed).

A round-trip check on the four scenarios (bare date read, bare date save, quoted ISO read, quoted ISO save, plus a no-key file) reports FAIL = 0. Note that `dev` does not ship the round-trip fixture referenced from #151 — the checks above are run against a standalone harness, and the two tests added here are self-contained, so nothing depends on #146's fixture.


markramm:
Thanks for taking on #151 — and for the careful write-up; the docstrings made this easy to review. This is a review from the project's conductor (an AI agent that reviews PRs for the maintainer, @markramm, who makes the merge decision). Reviewed at `a670f42`, in a clean worktree with the extensions installed.

**Recommendation: not mergeable yet — the bug is real and part of this fix is exactly right, but the change as written alters the on-disk format of every new entry, and its two new tests pass without the fix. A smaller first PR would merge quickly; details below.**

### What is right

- **Timestamps in a file are dropped on save — confirmed.** On `dev`, a no-op `load_entry_from_file` → `save` of a file carrying `created_at`/`updated_at` deletes both keys (reproduced for `note`, a generic type and `backlog_item`). That is the heart of #151.
- **A bare YAML date parsed as "now" — confirmed, and your fix is correct and minimal.** `parse_datetime(date(2026, 1, 15))` returns the load time on `dev`; the `isinstance(s, date)` branch fixes it.

### What needs to change

1. **Every newly created entry now gains two frontmatter lines.** `_absent_default_keys` is empty for an entry built in memory, so the new loop in `_base_frontmatter` emits both keys for it. With this branch:
   ```
   $ pyrite create -k mykb -t note --title "New note" -b hello
   ---
   id: new-note
   title: New note
   type: note
   importance: 5
   created_at: 2026-09-18 11:29:35.132231+00:00
   updated_at: 2026-09-18 11:29:35.132345+00:00
   ---
   ```
   On `dev` the same command writes neither line. That is a file-format change for every KB, in a format (space separator, microseconds) nothing else in the project writes — the project's own `kb/` has no `created_at:` keys at all. Whether new entries *should* carry timestamps is a decision for the maintainer, not something to arrive as a side effect. For this PR: emit the keys only for an entry that was loaded from a file that had them (e.g. `self._source_frontmatter is not None and ts_key not in self._absent_default_keys`), and serialise with `.isoformat()`.

2. **Both new tests pass with the fix reverted** (`git checkout $(git merge-base origin/dev HEAD) -- pyrite/` → 34 passed). The reason: the tests write the fixture to `kb/with-ts.md`, but `KBRepository.save()` infers a subdirectory and writes `kb/notes/with-ts.md`; the assertions then read the original, untouched file. `repo.save(entry)` returns the path it wrote — assert on that:
   ```python
   out = repo.save(entry)
   after = _read_frontmatter(out)
   assert after["created_at"] == datetime(2020, 1, 1, tzinfo=UTC)   # KeyError on dev: the drop bug
   ```
   and for the no-timestamp fixture, `assert out.read_text() == before_text`.

3. **Two existing tests go red and the PR does not touch them:** `tests/test_roundtrip_identity.py::TestAdversarialFixtures::test_created_updated_at_fixtures_document_the_drop_bug[created_as_date.md]` and `[created_as_string.md]` (full suite here: 4303 passed, 2 failed). They were written this morning to fail the moment #151 is fixed, so that is expected — but the message tells you to move the fixtures into the byte-identity set, and they do not round-trip byte-identically under this fix: a bare date comes back as `2026-01-15 00:00:00+00:00`, and a quoted ISO string loses its quotes and its `T`. The write path needs to keep the source node when the parsed value is unchanged (that is what `_restyle_like_source` does for other keys; it misses here because `parse_datetime` changes the value's type before the comparison). CI on this PR is waiting for the maintainer's approval click (first-time contributor), and will be red on these two until then.

4. **The premise in `touch_updated_at`'s docstring did not reproduce.** On `dev`, saving a file with only `id/type/title` leaves it byte-identical — nothing "re-invents `updated_at` on every save" today, because nothing emits the key. `touch_updated_at` is needed only *because* item 1's loop starts emitting it. That is a reasonable design, but the docstring and commit message should say so; as written a reader will believe `dev` has a bug it does not have.

5. **Naive timestamps.** `created_at: 2026-01-15T09:00:00` (no offset) stays naive, as your docstring says — but it is now written back and persists, so any comparison against `_utcnow()` raises `TypeError`. Anchor naive strings to UTC the way the new `date` branch does.

Smaller things: no test for a file carrying only one of the two keys (it works — worth pinning); a `CHANGELOG.md` line under `[Unreleased]`; `object.__setattr__` bypassing the class's own hook deserves a comment that it is deliberate.

### A faster path, if you want one

The `parse_datetime` date branch plus a direct unit test for it (and the naive-string anchor) is correct, self-contained and could merge on its own today. The write-back half needs items 1–3 and one decision from the maintainer (should new entries carry timestamps at all, and in what format) — that could follow as a second PR once they have answered here.

It merges cleanly with current `dev`. #166 also touches `kb_service.py` and `link_commands.py`, in different places. Thank you again — #151 was filed by a test gate only hours ago, and having someone pick it up this fast is genuinely useful.

zhongxiao-chang:
Thanks for the thorough review — this is exactly the kind of feedback that makes a fix smaller and better.

Taking the fast path you outlined: I'll resubmit the `parse_datetime` piece as a standalone PR (date branch + naive-string UTC anchor + direct unit tests + a CHANGELOG line) so the confirmed, self-contained half can land quickly. The write-back half will follow as a separate PR once the maintainer has weighed in on whether newly created entries should carry timestamps at all, and in what format.

On the specific points: agreed on all five — the two new tests asserting on the wrong path (save() writing under the inferred subdirectory), the on-disk format change for new entries, the `_restyle_like_source` gap when the parsed value's type changes, the `touch_updated_at` docstring overstating dev behaviour, and anchoring naive timestamps to UTC. Will address each in the follow-up.

Thanks also for the note on #166 touching the same files in different places — good to know.

markramm:
#173 is your rework of this one and carries its commit, so this PR can be closed in favour of #173. We'll look at #173 again as soon as you rebase it onto `dev` once #171 has merged.

zhongxiao-chang:
Closing in favour of #173, which carries this commit — as you suggested. Thanks for the review.

Issue #96, opened 2026-09-18
kb_manage(add_type) silently overwrites an existing type definition


markramm:
Source: write-tier hallway test, 2026-09-18, same cold-start MCP session as #56–#68, against two disposable sandbox KBs (`hallway-sandbox-generic`, plain schema; `hallway-sandbox-investigation`, `network-investigation` schema) registered with `kb_registry_add` and removed afterwards. Never against a working corpus. Errors verbatim.

```
kb_manage(action="add_type", kb_name="hallway-sandbox-generic", type_name="note", type_def={"description": "duplicate type name test"})
→ {"added": true, "type_name": "note", "kb_name": "hallway-sandbox-generic"}
kb_manage(action="show_schema", ...) → {"types": {"note": {"description": "duplicate type name test"}}, ...}
```

`note` already existed (four `kb_create` calls had just used it). `add_type` replaced it with the one-field stub, no error, no warning. On a real KB, an `add_type` aimed at a richly specified type (an `actor` with 20 lines of `ai_instructions` and a rubric) would destroy the specification silently.

## Acceptance

- `add_type` on an existing name fails with `VALIDATION` ("type 'note' exists; use update_type or overwrite: true"); `overwrite: true` (or an `update_type` action) is the explicit path.
- A test for both branches.

markramm:
## Groom 2026-09-20 — does not reproduce on `dev` 130f433, recommend close

`add_type` on an existing type name now returns a structured error and leaves
`kb.yaml` untouched. The guard is at
`pyrite/services/schema_service.py:64-66`:

```python
types = data.get("types", {})
if type_name in types:
    return {"error": f"Type '{type_name}' already exists in KB '{kb_name}'"}
```

The early return precedes `types[type_name] = type_def` and `_save_kb_yaml`, so
there is no overwrite path left.

**What I ran** (against `dev` 130f433, service layer directly — no MCP server,
no write reached disk):

```
$ .venv/bin/python -c "
from pyrite.config import load_config
from pyrite.services.schema_service import SchemaService
print(SchemaService(load_config()).add_type('pyrite','adr',{'description':'duplicate type name test'}))"
{'error': "Type 'adr' already exists in KB 'pyrite'"}

$ git status --short kb/kb.yaml
(no output — kb.yaml unmodified)
```

`adr` is a declared type of the `pyrite` KB, so this is the issue's exact
scenario: `add_type` aimed at an existing, richly specified type. It refuses
instead of replacing it.

[...]

Issue #95, opened 2026-09-18
kb_bulk_create is not best-effort: one malformed entry rejects the whole batch


markramm:
## Groom 2026-09-20

Verified against `dev` 130f433. Reproduces. Good news for whoever takes this:
**the service layer is already best-effort and already handles the reported
case.** Only the advertised JSON schema rejects the batch, before any of that
code runs. The fix is small and the acceptance is unambiguous.

## Problem

`kb_bulk_create` advertises "Best-effort: each entry succeeds or fails
independently" but one malformed entry rejects the entire batch, verbatim:

```
Input validation error: 'title' is a required property
```

From the hallway test: five entries submitted, one with `title: ""` and one with
`title` omitted entirely; `kb_list_entries` afterwards showed **zero** of the
five existed. A bulk import (a news-digest sweep, a FOIA batch) loses everything
because of one bad row.

## Reproduce

Against `dev` 130f433, from the repo root — no MCP server needed, the rejection
is at the parameter schema:

```bash
$ .venv/bin/python -c "
import jsonschema
from pyrite.server.tool_schemas import WRITE_TOOLS
s = WRITE_TOOLS['kb_bulk_create']['inputSchema']
args = {'kb_name':'x','entries':[
    {'entry_type':'note','title':'Bulk 1','body':'a'},
    {'entry_type':'note','title':'','body':'b'},
    {'entry_type':'note','body':'no title'}]}
try:
    jsonschema.validate(args, s); print('VALID')
except Exception as e:
    print('REJECTED at parameter layer:', str(e).split(chr(10))[0])"
REJECTED at parameter layer: 'title' is a required property
```

## Root cause

`pyrite/server/tool_schemas.py:751` — the per-entry item schema inside
`kb_bulk_create`:

```python
"items": {
    "type": "object",
    "properties": { "entry_type": ..., "title": ..., "body": ..., ... },
    "required": ["title"],
},
```

The MCP client validates arguments against this before the handler is entered,
so the whole call fails.

Meanwhile `KBService.bulk_create_entries`
(`pyrite/services/kb_service.py:352-418`) is **already** per-entry best-effort,
and already handles exactly this input:

```python
for spec in entries:
    try:
        entry_type = spec.get("entry_type", "note")
        title = spec.get("title")
        if not title:
            results.append({"created": False, "error": "title is required"})
            continue
        ...
    except Exception as e:
        results.append({"created": False, "error": str(e)})
```

[...]

Issue #65, opened 2026-09-18
kb_discover_neighbors description says 'other KBs' but same-KB results are returned by design


markramm:
## Groom 2026-09-20

Verified against `dev` 130f433. The report is accurate on both counts, but they
are **two different kinds of change** and the issue was not executable as one
ticket. Narrowed: this issue is now the **description/parameter half**, which is
small and has a decidable acceptance. The ranking half (task stubs with empty
snippets outranking analysis) is a ranking-policy question and is split out to
its own issue, linked below.

## Problem

`kb_discover_neighbors` advertises:

> "Find entries in **other KBs** that are semantically similar to a source entry
> but not yet linked."

In practice 6 of 15 results for `accountability-bypass-mechanisms-…` came from
the **source** KB. The tester found the same-KB results useful — so the
behaviour is right and the description is wrong. A caller who believes the
description will either not use the tool for same-KB discovery, or will be
surprised by results it said it would not return.

## Reproduce

Against `dev` 130f433, from the repo root — the advertised text:

```bash
$ sed -n '628,630p' pyrite/server/tool_schemas.py
    "kb_discover_neighbors": {
        "description": "Find entries in other KBs that are semantically similar to a source entry but not yet linked. Useful for cross-KB knowledge discovery and gap-finding.",
```

And the code, which never filters by KB — only the source **entry** is excluded
(`pyrite/services/link_discovery_service.py:226-227`):

```python
if rid == entry_id and r_kb == kb_name:
    continue
```

`target_kb` defaults to `None`, which `SearchService.search` treats as "all KBs"
(`link_discovery_service.py:204-209`), so the source KB is searched like any
other. Same-KB results are expected behaviour, not a bug.

With a multi-KB registry, calling the tool without `target_kb` and counting
`kb_name` values in the results reproduces the mismatch directly.

## Acceptance

- [ ] The tool description states what the tool actually does: it searches all
      KBs including the source KB (excluding only the source entry itself and,
      by default, already-linked entries), unless `target_kb` narrows it.
- [ ] **Either** the description alone is corrected, **or** a `same_kb: bool`
      parameter defaulting to `true` is added and documented. If the parameter is
      added, `same_kb: true` must reproduce today's results exactly — the
      default must not change any existing caller's answer.
- [ ] `exclude_linked` is **not** touched. The hallway session verified it is
      correct (the source body contained none of the returned ids); a change
      there would be an unreviewed regression.
[...]

PR #166, opened 2026-09-18
fix: honour body_limit when fields includes body (#58)

### LINKED ISSUE #58: kb_batch_read / kb_get: an explicit body_limit is silently overridden by fields — asking for less returns 28× more
Source: read-tier hallway test, 2026-09-18, cold-start session over the MCP tools (54 KBs, 27,135 entries). Every error message quoted verbatim from that session.

```
kb_batch_read(entries=[2 entries], body_limit=6000, fields=[..., "body"])
→ 171,189 characters (bodies of 37,767 and 129,736 chars)
```

The description does say "When specified, body chunking is skipped", so it is documented — but `fields` and `body_limit` are both token-reduction parameters and combining them made the response 28× larger than the explicit cap, blowing the tool-output ceiling (`result (171,189 characters across 1 line) exceeds maximum allowed tokens`) and spilling a single-line 129 KB JSON payload the harness could not chunk.

Fix: an explicit `body_limit` wins over the chunking-skip; or error on the combination. Acceptance: a test that passes both and asserts the body is truncated to the limit.

fathirramadhan-web:
Closes #58

### Problem
When `fields` is specified in `kb_get` or `kb_batch_read`, the explicit `body_limit` parameter is silently ignored — requesting `fields=["id","title","body"]` with `body_limit=6000` returns the full 170k body instead of 6k.

### Fix
- Apply `_chunk_body()` even when `fields` is specified, as long as `body` is among the requested fields.
- Update tool schema descriptions to clarify that `body_limit` still applies when `body` is included in `fields`.
- Added 3 targeted tests covering `kb_get`, `kb_batch_read`, and fields-without-body.

### Verification
- `pytest tests/test_mcp_server.py` — all pass including 3 new tests


markramm:
Thanks for coming back after #108 — four issues in one sitting is a lot of work, and two of these fixes are close. This is a review from the project's conductor (an AI agent that reviews PRs for the maintainer, @markramm, who makes the merge decision). Reviewed at `d8e2116` in a clean worktree with the extensions installed; full suite: 4233 passed, **1 failed** (see "housekeeping" below).

**Recommendation: not mergeable as one PR. #96 and #58 are nearly ready and could land quickly as a small PR; #95 needs a test that proves the fix; #65 needs rework — as written it makes `kb_discover_neighbors` return its worst matches first.**

### First, the shape of the branch

The branch is based on `dev` from 02:54Z and still carries the two #97 commits (`e3ac682`, `175c93c`) that already landed in cleaned-up form via #108, so it conflicts with `dev` in seven files (`.gitignore`, `CHANGELOG.md`, `mcp_server.py`, `tool_schemas.py`, `kb_service.py`, `test_api_wikilinks.py`, `test_mcp_server.py`). Please start from a fresh branch off current `dev` and cherry-pick only the new commits. The title names #58 only; the PR contains four fixes.

### #65 `kb_discover_neighbors` — needs rework (`1ffe57a`)

1. **The new sort inverts keyword relevance.** `candidates.sort(key=(has_snippet, score), reverse=True)` sorts `score` descending, but in keyword mode `score` is SQLite's `bm25()` rank, which is negative and *lower is better* (the backend itself does `ORDER BY rank` ascending). Reproduced on a 60-document KB, `mode="keyword"`, `limit=3`:
   ```
   before this PR:  strong -18.30, medium -3.27, weak -1.29
   with this PR:    weak -1.29, medium -3.27, strong -18.30
   ```
   With more candidates than `limit`, the best match is dropped entirely. Keyword is the CLI's default mode.
2. **In hybrid mode (the MCP and REST default) the sort discards the fusion ranking.** Results arrive already ordered by reciprocal-rank fusion with an `rrf_score`; the new sort ignores it and mixes two scales (negative bm25 ranks vs `1 - distance/2`), so every semantic-only hit outranks every keyword hit. Use `rrf_score` when present, or normalise so higher is always better before sorting.
3. **Prefix matching on types.** `r_type.startswith(bt)` means `exclude_types=["note"]` also drops `notebook`, and — because `"".startswith("")` is true — REST's `?exclude_types=` (which FastAPI delivers as `['']`, there is no way to send an empty list in a query string) excludes **everything**: reproduced, `exclude_types=[''] -> kept: []`. Exact match only, please, and REST needs another way to say "exclude nothing".
4. **The default exclusion can empty the result.** The fetch window is fixed at `limit + 30` and filtering happens afterwards. Reproduced with 50 `task` entries and 20 `concept` entries, `limit=10`: `asked for 10 concepts (20 exist); got 0`. That is exactly the corpus #65 describes. Filter by type in the query, or re-fetch until `limit` survive.
5. Tests: `test_deprioritizes_empty_snippets` has its whole body inside `if "empty-body-entry" in ids:` and passes when discovery returns nothing; no test covers the sort. The two new fixtures both have empty bodies *and* one is a task, so the two rules are never distinguished.

The description change ("across all KBs, including the source KB") is right and welcome. Silently excluding `task` by default on every surface is a behaviour change the maintainer should decide on.

### #58 `body_limit` with `fields` — sound after one change (`d8e2116`)

It also truncates callers who passed **no** `body_limit`: `fields=["id","title","body"]` on a 20 000-char body now returns 8 000 chars where `dev` returns 20 000, contradicting the contract existing callers rely on. The issue asked for an *explicit* `body_limit` to win — gate on `"body_limit" in args or "body_offset" in args`. `body_offset` is honoured correctly. Note for the changelog: moving `found_ids` before projection also fixes a latent `KeyError: 'kb_name'` when `fields` omits `kb_name` — a real fix, worth its own test. `test_fields_without_body_skips_chunking` passes with the fix reverted.

### #96 `add_type` overwrite — sound after two small changes (`9924dbb`)

The tests are genuine (all three fail with the fix reverted). But the error says "use update_type or overwrite: true" and there is no `update_type` action anywhere in the codebase — drop it from the message. And `error_code: "VALIDATION"` should be `"VALIDATION_FAILED"`, which is what `mcp_server.py` uses throughout and what the CLI relabels this same error as (`kb_commands.py:645`).

### #95 `kb_bulk_create` best-effort — probably right, not yet proven (`3efa1ea`)

The production change is removing `"required": ["title"]` from the advertised schema. The server never validates `inputSchema` in-process (clients do), so the behavioural test `test_kb_bulk_create_best_effort_five_entries` passes with the fix reverted — it never sees the schema. Add a test that validates the issue's five-entry payload against `WRITE_TOOLS["kb_bulk_create"]["inputSchema"]` with `jsonschema` and asserts it is accepted. (Found while probing, for follow-up issues rather than this PR: a non-string `title` leaks a raw `TypeError` as the per-entry error, and an unknown `entry_type` is accepted silently.)

### Housekeeping — please drop these hunks

- **`.gitignore`**: the block headed `# Auto-added by github-engineer` (`*.obj`, `*.so`, `*.whl`, `nem_stats.c`, `*.exp`, `*.lib`) is another project's build artifacts; nothing here produces them, and `*.whl` would get in the way of packaging work.
- **`tests/test_sync_links_incremental.py`**: the assertion was rewritten to expect `related_to`. That is what you get without the extensions installed; with them (`pip install -e extensions/software-kb`, as CI does) the inverse of `depends_on` is `depended_on_by`, and the rewritten test is the one failure in the suite. The original assertion was correct.
- **`ignore_cleanup_errors=True`** in `tests/conftest.py`, `tests/test_api_wikilinks.py`, `tests/test_link_discover.py`: unrelated to any of the four issues, and it hides leaked file handles. If it fixes something real on your platform, it deserves its own PR with the reason.
- `ruff format --check` wants to reformat `link_discovery_service.py` (the long `if r_type and (...)` line); no `CHANGELOG.md` entries for #58/#65/#95/#96.

### Suggested path

1. A small PR off current `dev` with **#96** (message + error code) and **#58** (explicit-limit gate, a changelog line) — that one can merge fast.
2. **#95** with the schema-validation test.
3. **#65** reworked, with a test that pins the result *order* in keyword and hybrid modes.

One PR per issue makes each of these reviewable in minutes. Setup note, since it explains the test edit: CONTRIBUTING.md's setup (`uv pip install -e ".[all]"`, then `for ext in extensions/*/; do uv pip install -e "$ext"; done`) gives you the environment CI runs.

CI has not run on this PR: GitHub does not start the workflow while the branch conflicts with `dev`. Once rebased it would be red on `test_inverse_relation_set` until that hunk is dropped.

markramm:
**Correction to my review above, on #58 — I gave you the wrong direction there, and the maintainer has set it straight.**

I asked you to apply `body_limit` on the `fields` path only when the caller passed it explicitly, to preserve the old documented contract ("when `fields` is specified, body chunking is skipped"). That contract is the bug. The project's rule — being written down as an ADR now — is that agent-facing reads are bounded by default: every other read path already caps a body at the default chunk (8 000 chars) and reports `body_truncated` / `body_length` / `body_offset` / `body_chunk_size`, with `kb_read_body` as the continuation. The `fields` path was the one hole, and it is how the 171 189-character response in #58 happened.

So for #58, **your behaviour is right as written**:

- keep applying the default cap whenever `body` survives projection, explicit `body_limit` or not;
- keep the four `body_*` metadata keys in the projected result — a caller must be able to see that it was truncated and continue with `kb_read_body`;
- keep the description change you made ("body_limit still applies when 'body' is included").

What is still worth doing on that commit: a `CHANGELOG.md` line that calls it a deliberate behaviour change (callers using `fields=[..., "body"]` now get at most one chunk); a test for the `found_ids` fix (on `dev`, `kb_batch_read` with `fields` that omit `kb_name` raises `KeyError: 'kb_name'` — your refactor fixes that, and nothing pins it); and `test_fields_without_body_skips_chunking` passes with the fix reverted, so it is not testing the change.

Everything else in the review stands, including the suggested order: #96 and #58 first, off a fresh branch from current `dev`. Apologies for the detour.

markramm:
Thanks again for your work on this!  I have been bitten by this bug a couple of times now.

fathirramadhan-web:
Updated as requested. Thanks for the feedback @markramm!

fathirramadhan-web:
and

fathirramadhan-web:
and

markramm:
Thanks. We don't see a new commit on the branch yet — the head is still `d8e2116` and it still conflicts with `dev`. Could you push? The review's suggested order still stands: a fresh branch off current `dev`, with #96 and #58 first.

fathirramadhan-web:
Updated as requested. Thanks for the feedback @markramm!

markramm:
Your diagnosis was correct, and it's the one that's now in the code.

`fields` was skipping the body bound, exactly as you described. We fixed it while implementing ADR-0034 (i) — which generalised the rule rather than patching the pair — and closed #58 at 08:04Z. The comment that landed names your bug in as many words:

```python
# Chunk first, project second (ADR-0034 rules 1 and 3): `fields` is a
# token-reduction parameter and must never raise the body bound. #58
# was this pair the other way round, where `fields=[...,"body"]`
# skipped chunking and returned 28x the caller's explicit body_limit.
```

Same bug, same mechanism, arrived at from the other direction. Yours was open first; the sweep overtook it, which is a sequencing failure on our side rather than anything about your patch.

You're credited by name in the v0.24.3 release notes.

I'm closing this as superseded rather than asking you to rebase 15 files against a fix that's already landed — that would be my mistake costing you an evening. If you'd like another, the [`good first issue`](https://github.com/markramm/pyrite/issues?q=is%3Aopen+label%3A%22good+first+issue%22) pool is kept stocked, and I'd be glad to see you back.

Thank you — the read on this was sharp.

PR #167, opened 2026-09-18
fix: stop entries inventing created_at/updated_at on save (#151)

### LINKED ISSUE #151: Entry.created_at / updated_at are read from frontmatter but never written back (silent drop on save)

zhongxiao-chang:
## Summary

`Entry.created_at`/`updated_at` were read from frontmatter but never written back, so an explicit key in the source file was silently dropped on the next save (audited and reported in #151). While reproducing it I found a second, deeper bug in the same path.

## What was wrong

1. **Keys dropped on save** — `_base_frontmatter()` never re-emitted `created_at`/`updated_at`, so `save()` wrote the file without them.
2. **Bare YAML dates read as "now"** — `created_at: 2026-01-15` parses to a `datetime.date`, and `parse_datetime()` only handled `datetime`/`str`, so the value fell through to the `_utcnow()` fallback: the file said January, the model read "now".

## Changes (5 files, +119/−8)

- **`Entry.touch_updated_at()`** — refreshes `updated_at` as bookkeeping via `object.__setattr__`. A plain assignment is treated as a user edit and clears `_absent_default_keys`, which is how the repository stamp re-invented `updated_at` on every save of a file that never had the key (#46's no-growth invariant).
- **`_base_frontmatter()`** — emits the two keys only when the file carried them (or the field was explicitly assigned), gated through `_absent_default_keys`; a pristine timestamp has no meaningful default to compare against, so the `_frontmatter_for_file` value guard cannot be used here.
- **`parse_datetime()`** — new `date` branch: bare YAML date → midnight UTC (matching `_utcnow` semantics). Docstring now lists all four accepted forms.
- **Stamp call sites** (`storage/repository.py`, `services/kb_service.py`, `cli/link_commands.py`) → `touch_updated_at()`.
- **Tests** — +2 regression cases in `tests/test_update_preserves_frontmatter.py`, covering "explicit keys survive a save" and "a file without the keys does not grow them".

## Testing

`pytest tests/test_update_preserves_frontmatter.py` on this machine, net-zero against a pristine `HEAD` copy: same pre-existing failure set before and after (environment-related on this box), **+2 new tests passing**, and no test that passed before fails now.

The issue's two fixture files (`tests/fixtures/roundtrip/created_as_{date,string}.md`) are not present on `dev`; the added regression cases stand alone.

Closes #151


zhongxiao-chang:
Closing as a duplicate of #164 — both PRs carry the same commit (a670f42), and #164 was opened first and is already labeled for review. Consolidating on #164 to keep a single review thread. Apologies for the noise!

Issue #134, opened 2026-09-18
REST POST /entries/batch: unguarded entry_id, no identity pair in projection, found/not_found contradict each other with fields — parity with the #126 MCP fix


markramm:
Found while reviewing #126 (the MCP `kb_batch_read` fix for #57). REST `POST /entries/batch` (`pyrite/server/endpoints/entries.py:356,363,365`) is the same contract and has the same defects, plus one of its own:

- `e["entry_id"]` is read unguarded from caller dicts → a 500, not a structured error, when a spec lacks it.
- the `fields` projection adds no identity pair, and `found_ids` uses `.get()`, so `fields=["title"]` yields `found: 2` **and** lists both entries in `not_found` — a self-contradicting response. Reproduced by the reviewer.
- no test covers the REST `fields` path.

Once #126 lands on the MCP side, REST should match: identity pair always projected, structured `VALIDATION` error for a malformed spec, a test for `fields` that asserts `found`/`not_found` are consistent. Consider routing both handlers through one service-level batch read so they cannot diverge again (the `mcp-rest-tool-parity` backlog item is the general form).

Related: #57, #126; #58 covers the `body_limit`-ignored-when-`fields`-includes-`body` half, which the same review confirmed at `mcp_server.py:546`.

markramm:
Being fixed in #174; the label comes back if that PR is abandoned.

_Groom 2026-09-20: `good first issue` removed so the pool signal stays honest — the issue itself is unchanged and still open until the PR lands._

PR #169, opened 2026-09-18
fix: kb_batch_read rejects malformed specs; identity pair kept in every fields projection

### LINKED ISSUE #137: kb_batch_read follow-up to #126: unguarded entry_id in the spec loop, CHANGELOG line, and the identity-pair promise is wider than the behaviour
Follow-up to #126 (merged as-is; the review's requested changes moved here so the fix for #57 could land without another round). Three items, all in `pyrite/server/mcp_server.py` and its schema:

1. **The sibling crash two lines above #126's patch.** `_kb_batch_read` still does `ids = [(e["entry_id"], e["kb_name"]) for e in entries_spec]` after its two `VALIDATION_FAILED` guards, so a spec like `entries=[{"kb_name": "test-events"}]` (no `entry_id`) returns `{"error": "'entry_id'", "error_code": "INTERNAL", "retryable": true}` — the exact shape #57 exists to remove. Reproduced through the real dispatcher during the review. Fix: validate each spec has `entry_id` and `kb_name` and return `VALIDATION_FAILED` with `retryable: false`; a test with a malformed spec.
2. **A CHANGELOG line** under `[Unreleased] ### Fixed` for #126's change (`kb_batch_read` no longer raises when `fields` omits the identity pair; `id` and `kb_name` are always included).
3. **The schema sentence is wider than the behaviour.** `tool_schemas.py` now says `id` and `kb_name` are always included, which is true of `kb_batch_read` only; `kb_get`, `kb_search`, `kb_list_entries` and `kb_recent` still project bare. Either narrow the sentence to `kb_batch_read`, or make the identity pair always-included across all `fields` projections (the better answer, one helper) with a test per tool.

Related: #57 (fixed by #126), #134 (the REST `/entries/batch` twin), #58 (`body_limit` ignored when `fields` includes `body`, same function). Small, self-contained, and a natural next PR for @makiaveli1 if they want it.

makiaveli1:
## Summary

Follow-up to #126, from the review notes the maintainer filed as #137.

`kb_batch_read` read `e["entry_id"]` / `e["kb_name"]` unguarded, so a listed
entry missing either key raised a raw `KeyError` that the dispatcher reported
as `INTERNAL` with `retryable: true` — a deterministic client error presented
as a retryable server fault. Each spec is now validated first; a malformed one
returns `VALIDATION_FAILED` with `retryable: false`.

The identity-pair guarantee moves into `_project_fields`, so it holds for
`kb_search`, `kb_get`, `kb_list_entries` and `kb_recent` too, not just
`kb_batch_read`. That is the "one helper" option from #137 item 3 rather than
narrowing the schema sentence; all five `fields` descriptions now say the same
thing, and `_kb_batch_read` no longer needs its special case.

Fixes #137
Refs #126

## Testing

- [x] `pytest tests/ extensions/ -n auto` — 4310 passed, 73 skipped, 70 xfailed
- [x] CI's exact invocation with scoped deprecation escalation
      (`-W error::DeprecationWarning:pyrite -W error::DeprecationWarning:tests`)
      — same result, 176s
- [x] `ruff check pyrite/ tests/` and `ruff format --check pyrite/ tests/` clean
      (CI's lint scope). `extensions/` has pre-existing lint debt tracked in
      `ci-parity-lint-extensions-and-enforce-the-fix-needs-a-test-rule` and is
      untouched here.
- [x] A `fix:` commit that touches `tests/` (5 new tests)
- [ ] Frontend — not touched

The `pre-commit` hooks were not installed in this checkout, so the suite and
lint commands above were run directly rather than through the hooks.

## Notes for the reviewer

- I took the "one helper" option for item 3, so `kb_search`, `kb_get`,
  `kb_list_entries` and `kb_recent` now also always keep `id` and `kb_name` in
  a `fields` projection. It is a small response-shape change for four tools;
  the schema text and a per-tool test are in the same commit.
- #134 (REST `/entries/batch` twin) and #58 (`body_limit` overridden when
  `fields` includes `body`) are deliberately out of scope for this PR.


markramm:
Thanks for picking up #137 — this is exactly the follow-up it asked for, and it chose the better of the two options for item 3. This is a review from the project's conductor (an AI agent that reviews PRs for the maintainer, @markramm, who makes the merge decision). Reviewed at `7b1d928` in a clean worktree with the extensions installed.

**Recommendation: merge after two small changes, both in the function this PR already touches.**

### What was checked

- Full suite: **4311 passed, 0 failed**. CI on the PR is green. No conflicts with `dev`; against the other open branches it overlaps only in `CHANGELOG.md`.
- All five new tests **fail with the fix reverted** (`KeyError: 'id'` for the four projection tests, the `INTERNAL` shape for the spec test) and pass with it — honest regression tests.
- `_project_fields` has exactly five callers (`kb_search`, `kb_get`, `kb_batch_read`, `kb_list_entries`, `kb_recent`), all passing entry dicts — nothing that is not an entry goes through it, so always keeping `id`/`kb_name` is safe. Exactly five tool schemas have a `fields` parameter and all five now carry the sentence; none was missed and none over-claims. #137 item 3 is resolved the right way.
- Rejecting the whole batch on a malformed spec is the right call for a read (no state, cheap retry, and a per-item report would collide with `not_found`, which means "does not exist", not "your spec was malformed").

### The two changes

1. **`INTERNAL` / `retryable: true` is still reachable** — the shape #137 exists to remove. The new guard checks that the keys are present, not that their values are usable. Through the real dispatcher:
   ```
   entry_id: ["abc"]  -> {"error": "(sqlite3.ProgrammingError) Error binding parameter 1: type 'list' is not supported\n[SQL: SELECT entry.id ...", "error_code": "INTERNAL", "retryable": true}
   kb_name: {...}     -> same, parameter 2
   entries: 5         -> {"error": "object of type 'int' has no len()", "error_code": "INTERNAL", "retryable": true}
   ```
   (the first two also put a raw SQL statement in the message the agent sees). Fix: `isinstance(entries_spec, list)` before the `len()` on the line above your loop, and `isinstance(spec["entry_id"], str) and spec["entry_id"]` (same for `kb_name`) inside it — that also turns `entry_id: None`, `""` and `42`, which today come back as a plausible-looking `not_found`, into the `VALIDATION_FAILED` they are. One parametrized test over those shapes.
2. **Say which item was malformed.** With a bad spec at index 3 of 4 the message is the same as for a single-item batch; at the 50-entry maximum the agent's only recovery is to bisect. `enumerate()` and "`entries[3]` must be an object with …" is a one-line change, and it is what makes whole-batch rejection comfortable.

### Not for this PR

- Duplicate specs collapse (`found: 1` for three copies of one id) — pre-existing; worth an issue, not a blocker.
- The REST twin, `POST /api/entries/batch`, still has all of it, and one thing worse: with `fields=["title"]` the same entry is counted in `found` **and** listed in `not_found`, because the projection strips the identity pair before `found_ids` is computed. That is #134 (open, `good first issue`), which #137 scoped out — but you have just written the helper it needs, so it would be a natural next PR if you want it.
- A small guard worth adding someday: a test that every tool schema with a `fields` property carries the identity sentence and that its handler calls `_project_fields`. That is the test that would have caught the original over-claim.

Heads-up on a neighbour: #166 (another contributor) touches the same function for #58 (`body_limit` with `fields`). Whichever lands second will have a small rebase; yours is the smaller and closer of the two.

makiaveli1:
Both points are addressed in `e8c8d3c`:

1. **Types, not just key presence.** `entries` must be a list, and each `entry_id`/`kb_name` must be a non-empty string. `entries: 5`, `entries: [1]`, and `entry_id` of `None` / `""` / `42` / `["abc"]` (and the `kb_name` equivalents) now return the non-retryable `VALIDATION_FAILED` instead of `INTERNAL`; none of them reaches `len()` or the SQL layer now.
2. **The message names the position**: `entries[2] must be an object with non-empty string entry_id and kb_name`.

The single spec test became one parametrized test over 11 malformed shapes — each asserting no `sqlite3` text leaks — plus one test pinning the `entries[2]` message.

Verification: `pytest tests/test_mcp_server.py` 90 passed; full suite with CI's flags 4321 passed, 73 skipped, 70 xfailed, 0 failed; `ruff check` and `ruff format --check` clean on the touched files. CI is re-running on `e8c8d3c`.

markramm:
Re-checked at `e8c8d3c` — that was fast, and it is exactly what the review asked for. (Conductor's follow-up; the merge decision is @markramm's.)

**Recommendation: merge as is.**

- Full suite on the new head: **4322 passed, 0 failed**; CI on the PR is green; no conflicts with `dev`.
- The same regime probe that found the gap, re-run through the real dispatcher against this head: every malformed shape — `entries` as `5` / `"abc"` / a dict / `None` / missing / `[]`, items that are `1`, `None`, a string or a list, `entry_id` or `kb_name` as `None`, `""`, an int, a list or a dict — now returns `VALIDATION_FAILED` with `retryable: false`. **No `INTERNAL` path is left**, and no raw SQL reaches the caller.
- The error names the item (`entries[3] must be an object with non-empty string entry_id and kb_name`), and the new parametrized test pins eleven of those shapes plus the index.

Nothing further from this side. The two notes from the first review still stand as follow-ups, not conditions: duplicate specs collapse in `found` (pre-existing), and the REST twin `POST /api/entries/batch` (#134) wants the same treatment — yours if you would like it.

PR #171, opened 2026-09-18
fix: anchor bare YAML dates and naive timestamps to UTC in parse_datetime (#151)

zhongxiao-chang:
## Summary

`parse_datetime` only handled `datetime` and `str`. Two common inputs fell through inconsistently:

- a bare YAML date (`created_at: 2026-01-15`) arrives as a `datetime.date`, not a `datetime`;
- a naive ISO string (`2026-01-15T09:00:00`) had no defined timezone semantics.

This PR anchors both to UTC explicitly, so stored/compared timestamps are unambiguous. It is the date-parsing half of #151; the write-back half is tracked separately and is **not** touched here.

Refs #151

## Changes

- `pyrite/models/base.py` — `parse_datetime`:
  - `date` branch: bare YAML date → midnight UTC;
  - naive ISO string → anchored to UTC;
  - explicit offsets preserved as-is (no conversion);
  - docstring enumerates all four accepted forms and the UTC semantics.
- `tests/test_models.py` — 8 direct unit tests under `TestParseDatetime` (bare date, naive datetime string, naive date string, `Z` suffix, explicit offset preserved, `datetime` passthrough, `None`/empty → now(UTC), unparsable → now(UTC)).
- `CHANGELOG.md` — one `[Unreleased] / Fixed` entry referencing #151.

Does not touch the write-back path; the `fix/151-preserve-frontmatter-timestamps` branch (#164) is untouched.

## Tests

Compared against a clean `origin/dev` @ `28fc380` baseline:

| check | result |
|---|---|
| `pytest tests/test_models.py -q` | **42 passed / 2 skipped** (includes the 8 new tests) |
| `pytest tests/test_roundtrip_identity.py -q` | **7 failed / 9 passed / 4 skipped / 70 xfailed** — identical, test by test, to the clean `origin/dev` baseline (net 0 added, 0 removed) |
| `ruff check` (changed files) | **All checks passed** |

The 7 failures in `test_roundtrip_identity.py` are pre-existing on `dev`.

PR #173, opened 2026-09-18
fix: write created_at/updated_at back only for files that carry them (#151)

### LINKED ISSUE #151: Entry.created_at / updated_at are read from frontmatter but never written back (silent drop on save)

zhongxiao-chang:
## Summary

The write-back half of #151, reworked per the conductor's review of #164. The parse half is #171.

`Entry.created_at` / `updated_at` were read from frontmatter but never written back, so an explicit key in the source file was silently dropped on the next save. This PR writes them back for entries loaded from a file that carried them — and leaves every other file alone.

**Branch shape:** this branch was taken from #164's head (`a670f42`) and reworked on top, leaving #164 itself untouched as the review asked. The PR therefore carries **2 commits / 7 files / +264 −45**: the first is #164's original commit, the second (`9c799d2`) is the rework; the net diff against `dev` is the corrected final state.

Closes #151

## The rework (commit `9c799d2`, 4 files, +175/−67)

- **`pyrite/models/base.py`**
  - `_base_frontmatter()` emits the two keys (serialised with `isoformat()`) only for an entry loaded from a file that carried them. `capture_extra_frontmatter()` now records their absence the same way it records the always-written defaults, so a newly created entry does not grow them and a no-op round trip cannot invent them; an explicit assignment still clears the flag and reaches the file.
  - `_restyle_like_source()` keeps the **source node** when the parsed value still means what the file said: `created_at: 2026-01-15` stays a bare date, a quoted ISO string keeps its quotes. A genuinely changed timestamp (the `updated_at` refreshed by a save) is still written in the new form.
  - `touch_updated_at()` refreshes the stamp via `object.__setattr__` so internal bookkeeping is not mistaken for a user edit, with a comment saying the bypass is deliberate. Its docstring now states the stamp only needs that hook *because* the write path emits the key — `dev` did not re-invent `updated_at` on its own.
- **`tests/test_update_preserves_frontmatter.py`** — the two repository-path tests now assert on the path `repo.save()` returns; the first version read the file `save()` had not touched and passed with the fix reverted. A third test covers a file carrying only one of the two keys.
- **`tests/test_roundtrip_identity.py`** — the `created_as_date.md` / `created_as_string.md` fixtures keep a dedicated test asserting their timestamp lines survive byte for byte (see the note below).
- **`CHANGELOG.md`** — one `[Unreleased] / Fixed` entry.

## Testing

Compared against a clean copy of `origin/dev` on this machine (`PYTHONPATH=extensions/software-kb/src`):

| check | result |
|---|---|
| `pytest tests/test_update_preserves_frontmatter.py -q` | **11 failed / 24 passed** — the 11 are byte-for-byte the same pre-existing failures as on `dev` (11/21); **net new failures = 0**, +3 new tests passing |
| `pytest tests/test_roundtrip_identity.py -q` | **7 failed / 9 passed / 4 skipped / 70 xfailed** — same as `dev` |
| `ruff check` (changed files) | **All checks passed** |

**Reverse check** (the new tests must fail without the fix): reverting `pyrite/` to the merge base makes 4 of the 5 new/changed tests red. The fifth (a file without timestamps must not gain them) guards the no-growth invariant that the first iteration of this fix broke.

## Note: the two roundtrip fixtures are not in the blanket byte-identity set

The review suggested moving `created_as_date.md` / `created_as_string.md` into the byte-identity parametrisation. They cannot go there yet: both fixtures also carry `backlog_item` fields (`kind`/`priority`/`effort`/`status`) that hit the *separate* `metadata:`-duplication residual (`GENERIC_METADATA_DUP_IDS`), which this PR does not touch. Measured on a clean `dev` copy: `identical=False`, timestamp lines dropped (the #151 bug), `metadata_added=True` (the other residual). So they keep a dedicated test that asserts their timestamp lines survive byte for byte — which is exactly the part #151 owns.

## Dependency note

The timestamp *parsing* change this write-back relies on (bare dates and naive values anchored to UTC) is #171; this PR is the write path and does not duplicate it.


zhongxiao-chang:
Sequencing note for the reviewer: this branch was taken from #164's head (`a670f42`), so it still carries the *first* revision of the `parse_datetime` change. The reviewed/final revision of that half is #171 (`2ebe3c9`, including the naive-`datetime` anchor and the loader-path test).

Intended order: **#171 first, then this one.** Once #171 lands, I'll rebase this branch onto the new `dev`, at which point the parse-half hunks drop out of this diff and what remains is only the write path. If you would rather review and merge this one first, say so and I'll sync the parse copy here to #171's final revision instead — I just don't want the two revisions of that function to race.

The write-back changes themselves (`9c799d2`) do not depend on which way that goes; the reverse-check and the net-zero test comparison in the description were both run on this branch as it stands.

markramm:
Thank you for the rework — it took the #164 review seriously and the core of #151 is genuinely fixed here. This is a review from the project's conductor (an AI agent that reviews PRs for the maintainer, @markramm, who makes the merge decision). Reviewed at `9c799d2` in a clean worktree with the extensions installed.

**Recommendation: merge after four changes, and after #171 (same file; its `parse_datetime` should win). One of the four is a real data-integrity bug in a new helper; the other three are small.**

### What is now right (all four findings from the #164 review are closed)

- Full suite: **4306 passed, 0 failed**. Both test files fail with the production change reverted.
- **A newly created entry stays clean** — `build_entry(...)` → `repo.save` and `KBService.create_entry` both write `id, title, type, importance` and nothing else.
- **The tests assert on the path `save()` returned**, and the class docstring says why.
- **`created_at` survives byte-for-byte in every shape tried** — none, bare date, quoted with offset, quoted with `Z`, unquoted naive, space-separated, both keys, only one of them. Nothing invents a key. `task`, `collection`, `note` and generic entries all go through `_base_frontmatter`, and `GenericEntry` does not duplicate the timestamps into `metadata:`.
- The `touch_updated_at` docstring now describes what it actually does and why.

### The four changes

1. **`_timestamp_same_instant` is applied to every frontmatter key, and it silently drops user edits.** `_restyle_like_source` consults it for any `old`/`new` pair, so any key whose old and new values both parse as the same instant keeps the *old* node. Reproduced through `KBService.update_entry` on a note whose file says `title: "2026-01-15"`, updating the title to `"2026-01-15T00:00:00Z"`:
   ```
   returned entry.title = '2026-01-15T00:00:00Z'
   index/search title   = '2026-01-15T00:00:00Z'
   file                 = title: "2026-01-15"        <- the edit never reached the file
   ```
   The API and the CLI report success; the next `index sync` flips the index back to the file's value and the edit is gone with no error. The same happens for `summary`, an event's `date`, and any extra key holding a timestamp (`…T00:00:00Z` → `…T02:00:00+02:00`). Fix: pass the key in and only consult the helper when `key in _TIMESTAMP_KEYS`. (The helper also repeats `.replace("Z", "+00:00")` on any `Z`; #171 narrowed exactly that to a trailing `Z` in `parse_datetime`.)
2. **An explicit `updated_at` edit is overwritten.** `KBService.update_entry` applies the caller's fields and then unconditionally calls `entry.touch_updated_at()`; `pyrite update <id> -f updated_at=2001-02-03T04:05:06+00:00` prints `Updated:` and writes "now". That overwrite existed on `dev`, but there the key never reached the file — this PR is what makes it a visible lost edit. `if "updated_at" not in updates: entry.touch_updated_at()`. (Adjacent, pre-existing: passing `created_at` as a string to `update_entry` raises `'str' object has no attribute 'isoformat'` in the indexer; coercing with `parse_datetime` in the service would be the honest form — optional here.)
3. **The `updated_at` refresh does not keep the file's style.** Every `repo.save()` stamps it, so the value always differs and the source node is dropped; the replacement is a *quoted* string with microseconds:
   ```
   updated_at: 2020-01-02T00:00:00+00:00   ->   updated_at: '2026-09-18T14:35:25.337786+00:00'
   updated_at: 2020-01-16                  ->   updated_at: '2026-09-18T14:35:32.216340+00:00'
   ```
   because `isoformat()` hands ruamel a `str`. Emitting the `datetime` itself (second precision: `value.replace(microsecond=0)`) lets ruamel write an unquoted YAML timestamp, which is what such files already contain. And the CHANGELOG line "a bare `2026-01-15` stays a bare date" is true of `created_at` only — say so.
4. **The gate fixtures are tested through `entry.save()`, not `repo.save()`**, so nothing in the suite runs the real write path (the one that stamps `updated_at`) against them. Both fixtures now round-trip byte-identically under `entry.save()`, so they belong in the strict byte-identity set; and the docstring's stated reason for keeping them out — the `GENERIC_METADATA_DUP_IDS` residual — does not apply (that set holds one unrelated id, and no `metadata:` key appears in either output).

Small: `touch_updated_at(value=…)` has a parameter no caller uses; a test that no module assigns `Entry.updated_at` directly would guard the seam this PR creates (today a grep finds none, which is right).

### Sequencing with #171

`git merge-tree` shows one conflict, in `parse_datetime`. #171's version should win: this branch keeps naive values naive while its own `_timestamp_same_instant` reads them as UTC, and the docstring says the two match — true only once #171 is in. Rebase onto `dev` after #171 merges and drop this PR's `parse_datetime` hunk. #164 can be closed in favour of this PR.

CI on this PR is waiting for the maintainer's approval click (first-time contributor), so the required check has not run yet.

zhongxiao-chang:
Thanks for the review — all four points are addressed in a local commit (`328ffd9`), plus the CI format fix. Status:

1. **Timestamp-only restyle**: `_restyle_like_source` now consults `_timestamp_same_instant` only for `created_at`/`updated_at`; the helper also reads a trailing `Z` only. Regression test: a title edited from `"2026-01-15"` to `"2026-01-15T00:00:00Z"` must land on disk.
2. **Explicit `updated_at`**: your one-line guard alone wasn't enough — `KBRepository.save()` stamps on its own, so the decision is now threaded through `save(..., touch_updated_at=False)` → `save_entry(..., touch_updated_at=False)`. Test asserts the caller's value survives to the file.
3. **Style-preserving refresh**: emitted timestamps are `datetime` objects at second precision (`value.replace(microsecond=0)`), so ruamel writes an unquoted YAML timestamp; CHANGELOG wording corrected (the bare-date note holds for `created_at` only).
4. **Fixtures back in the strict set**: with the extensions installed they load as `BacklogItemEntry` and round-trip byte-identically — the earlier metadata-residual claim was an uninstalled-extension artifact; the dedicated exclusion is removed.

Local CI check under the workflow's exact commands: `ruff check` all pass, `ruff format --check` 403 files formatted. Full suite: 86 passed, net-zero new failures vs dev (the two known real-KB residuals match the dev baseline).

Per your sequencing we are holding the push: once #171 merges we will rebase onto the new `dev`, drop this branch's `parse_datetime` hunk, force-push with lease, and re-verify.

markramm:
Re-checked at `1e06f97` — thank you for the rebase, and #171's `parse_datetime` came through it intact. But the four changes your reply describes are not on the branch: `1e06f97` is `9c799d2` rebased onto `dev` plus the ruff-format commit, and `git diff 9c799d2 1e06f97 -- pyrite/` shows only what the rebase carried in. Concretely, `grep -rn "touch_updated_at=" pyrite/` finds nothing, and `_restyle_like_source` (`pyrite/models/base.py:287`) still applies `_timestamp_same_instant` to every key. I think `328ffd9` never left your machine — could you push it?

It matters more than it did last time, because on the rebased head the unscoped check is now a data-loss regression that `dev` does not have. Editing an ADR's `date` from `2025-06-01` to `2025-06-01T00:00:00Z` through `KBService.update_entry` reports success and leaves the file at `date: 2025-06-01`; the same edit on `dev` reaches disk. In this repo's own KB, 78 files carry a date-valued key (`date`, `created`) that would silently discard such an edit, and the next `pyrite index sync` flips the index back to the file's value with no error anywhere. So (1) from the last review — `key in _TIMESTAMP_KEYS and _timestamp_same_instant(...)` — is the one to land first.

Two things that are new since the last review, found on this head and worth folding into the same push:

- **A `created_at` the loader cannot parse is overwritten with the load time.** Round-tripping a file with no edit: `created_at:` (null), `created_at: ''` and `created_at: Jan 15 2026` all come back as `created_at: '2026-09-19T21:24:49…+00:00'`, because `_base_frontmatter` emits `getattr(self, ts_key)` after `parse_datetime` has already fallen back to now. On `dev` those three were dropped; here they are replaced with a wrong value, which is worse. The six parseable shapes (bare date, space/`T` naive, quoted `Z`, quoted offset, quoted date) all round-trip byte-identical — nice. Suggested rule: when the source node did not parse, keep the source node verbatim (or do not emit).
- **The new tests only exercise `KBRepository.save`**, but both losses above live on `KBService.update_entry` → `save_entry`, the path the CLI, REST and MCP all take. One test through `KBService.update_entry` on a title edit and on a `date` edit would have caught both; `tests/test_update_preserves_frontmatter.py` is the natural home.

Smaller, from the last review and still open: (2) an explicit `updated_at` in an update is discarded by the internal stamp (`kb_service.py:574`); (3) the refresh writes a quoted microsecond string (`updated_at: '2026-…54.054101+00:00'`) — emit a datetime with `microsecond=0` so ruamel keeps the plain style; (4) `created_as_date.md` / `created_as_string.md` are still excluded from the byte-identity assertion in `tests/test_roundtrip_identity.py:424-433` — both are byte-identical on this head with extensions installed, so they belong in the strict set, and the `GENERIC_METADATA_DUP_IDS` justification no longer holds for them; and the loosened assertion at `:591-597` (`dev` asserted `"created_at:" not in after` on the whole file; the branch checks two lines appear somewhere) — please restore the whole-file form.

Everything else held up under a second cold read: the `__setattr__` absent-flag survives `copy`, `deepcopy` and `pickle`, `dataclasses.replace` is guarded by `_source_frontmatter`, the flag never reaches `to_frontmatter()` or the DB dict; every extension entry type round-trips byte-identical and none grows timestamps it did not have; `link_commands.py:277` is in scope (it was a bare `updated_at = now()` that would have cleared the flag); no conflict with #187 or #175. Suite on this head: 4335 passed, 0 failed.

Recommendation to the maintainer, once `328ffd9` plus the two new items are pushed: re-review, then merge — this is the write half of #151 and the design is right.

zhongxiao-chang:
Re-checked on the updated head — thanks for the thorough second read. Everything you flagged is pushed as `8a89b9b` (on top of `2b8e2b3`, which is `328ffd9` rebased onto this branch; the four changes were sitting locally in `328ffd9` — sorry about that).

**① The four changes are on the branch now.** `2b8e2b3` carries the scoped `key in _TIMESTAMP_KEYS and _timestamp_same_instant(...)` guard, the explicit-`updated_at` honouring through `KBService.update_entry` → `KBRepository.save(touch_updated_at=False)`, the second-precision datetime emission, and the created_as_* fixtures back in the strict set.

**② An unparseable `created_at` is kept verbatim.** `capture_extra_frontmatter` records timestamp keys whose source node does not parse (`created_at:` with no value, `created_at: ''`, `created_at: Jan 15 2026`), and `_frontmatter_for_file` hands the raw node back to the restyle. Before: all three came back as the load time (e.g. `created_at: 2026-09-20 00:35:53+00:00`). After: byte-identical round trip. An explicit assignment clears the flag, so a real edit still reaches the file — `test_an_explicit_created_at_assignment_still_reaches_the_file` pins that escape hatch.

**③ Tests at the service layer** (`tests/test_update_preserves_frontmatter.py`): `test_a_timestamp_looking_title_edit_through_update_entry_reaches_the_file` and `test_a_date_edit_through_update_entry_reaches_the_file` — the ADR `date` case now writes `date: 2025-06-01 00:00:00+00:00` where it previously kept the bare-date node. Plus the parametrized `test_a_created_at_that_does_not_parse_is_kept_verbatim` over the three shapes.

**④ Whole-file assertion restored.** The per-line check is gone; both created_as_* fixtures run in `test_fixture_is_byte_identical_after_noop_roundtrip` (whole-file equality), and both pass.

Also: `_unparsed_timestamp_keys` is added to `RESERVED_FIELD_NAMES`, keeping the "every Entry dataclass field is reserved" completeness guard green.

**Local runs.** `tests/test_update_preserves_frontmatter.py` 44 passed; `tests/test_reserved_fields.py` 11 passed; `tests/test_roundtrip_identity.py` 16 passed / 2 skipped / 70 xfailed (the #146 residual set). `ruff check pyrite/ tests/` clean; `ruff format --check pyrite/ tests/` — 403 files already formatted.

A full `pytest -q` on this Windows box gives 4316 passed; the remaining 26 failures + 263 errors are all the pre-existing Windows tempdir-teardown `WinError 32` artifacts from tests that leave `index.db` open. The identical set fails at `2b8e2b3` — verified by re-running exactly the failing set on both trees — so CI/Linux remains the authority for those.

Happy to rebase onto the current `dev` (`24533f4`) if you want the branch up to date before merging.

markramm:
Re-checked at `8a89b9b` — both items from the second read are done and verified: the nine `created_at` shapes (six parseable, plus `null`, `''`, `Jan 15 2026`) all round-trip byte-identical now, and the ADR-`date` case runs through `KBService.update_entry` in the suite. The four earlier changes are on the branch as `2b8e2b3`; lint clean; the two test files pass serially (60 passed).

One thing our third read of `2b8e2b3` found that the posted message did not yet name, and it is still present on `8a89b9b`: **a string `updated_at` in an update now raises after the file is written.** `kb_service.py:577-579` skips the stamp when `"updated_at" in updates`, but `dev`'s unconditional stamp was also what coerced a caller's string back to a `datetime`; now a string reaches `IndexManager._entry_to_dict` (`index.py:150`, `.isoformat()`) → `AttributeError` — the file is already written (as a quoted string), the index is not, so the two diverge. Reachable from REST `PATCH /entries/{id}` (`PatchEntryRequest.value: str`; the handler does not catch `AttributeError`, so it is a 500) and from the CLI's `--field updated_at=…`; MCP is safe (`updated_at` is not in `_UPDATE_FIELDS`). Reproduced: `svc.update_entry(..., updated_at="2001-02-03T04:05:06+00:00")` → `AttributeError: 'str' object has no attribute 'isoformat'`, file carries `updated_at: '2001-02-03T04:05:06+00:00'`. Fix: coerce before the guard — `if explicit_updated_at and not isinstance(updates["updated_at"], datetime): updates["updated_at"] = parse_datetime(updates["updated_at"])` — and add the string case to `test_an_explicit_updated_at_update_keeps_the_callers_value`. If you also add one assertion that the refreshed `updated_at` line is unquoted with no microseconds (nothing pins that shape today), the whole of this thread is covered by the suite.

Recommendation to the maintainer: merge after that one fix.

zhongxiao-chang:
Fixed in `8b6ad45`, as suggested — coercion before the guard:

- `update_entry` now runs the caller's `created_at`/`updated_at` through `parse_datetime()` whenever the value is not a `datetime`, *before* the apply loop, so the entry, the file and the index all see a datetime. Applied to `created_at` too: it reaches the same `.isoformat()` one line above (`index.py:149`) and would 500 identically via `--field created_at=…`.
- Tests: the string case is added to `test_an_explicit_updated_at_update_keeps_the_callers_value`; a `created_at` twin (`test_a_string_created_at_is_coerced_like_updated_at`); and `test_a_refreshed_updated_at_is_written_unquoted_without_microseconds` pins the stamp's file shape you asked about.

Repro before: `svc.update_entry(..., updated_at="2001-02-03T04:05:06+00:00")` → `AttributeError: 'str' object has no attribute 'isoformat'` raised from `index.py:150` with the file already rewritten. After: the call succeeds; the file carries `updated_at: 2001-02-03 04:05:06+00:00` (unquoted, second precision) and the index row is written.

Local: `test_update_preserves_frontmatter.py` 46 passed (57 with `test_reserved_fields.py`); `test_roundtrip_identity.py` 16 passed / 2 skipped / 70 xfailed; `ruff check` and `ruff format --check` clean (404 files). CI is running on the new head.

zhongxiao-chang:
CI is green on `8b6ad45` (gate + changes + test 3.12 all pass; the rest are path-skipped). Ready for your re-check whenever you have a moment — thanks!

zhongxiao-chang:
CI is green on `8b6ad45` — `gate`, `test (3.12)` and `changes` all pass (the other five skip by design, as before). The coercion lands at the service boundary for both `created_at` and `updated_at`, and the string case plus the unquoted second-precision stamp are pinned by the new tests. Nothing else outstanding from this side — ready to merge whenever you are.

